PR #284, opened 2015-12-15
Fixed UserPicture recycling issues

dlew:
I noticed that scrolling through the party members list would cause the pictures to jump around from one to the next. I've implemented some basic fixes for the problem that properly cancels existing loads when a new one comes in.


vIiRuS:
Thanks for the fixes!

I noted this as a contribution towards your second contributor tier!

Issue #287, opened 2015-12-15
Memory leaks abound in the app


dlew:
I tried installing [LeakCanary](https://github.com/square/leakcanary) to see what sort of memory leaks were occurring and found a whole ton of them. See what I got done before giving up: https://github.com/dlew/habitrpg-android/tree/dlew/memory-leaks

(I could submit this as a PR, but it won't help much since `MainActivity` will still leak all the time.)

There seems to be some fundamental issues with memory retention in the app that may require a more thorough re-examining of architecture. For example, `EventBus.register()` must be matched by a subsequent `EventBus.unregister()`, otherwise the object is leaked forever. I'm not sure I want to tackle this without knowing more of what you have in mind for `EventBus` (or the use of statics).

vIiRuS:
Thanks for bringing this to our attention! As to not lose the work you already did, a PR with the existing commits would be handy. I will then investigate the issue further.

christopherfowers:
What is the status of this? The pasted link is broken now. If like to start looking into this issue to see what I can contribute. Has there been any recent analysis done? Just trying to not duplicate work here.

saraolson:
Since a lot has changed since this issue, we're gonna close it and if any memory leaks are still happening, let a new issue be opened with more specific details.

PR #303, opened 2015-12-16
Added LeakCanary and began fixing memory leaks

dlew:
This begins to address #287, though there are plenty more memory leaks out there.

Warning: This will add LeakCanary to debug builds, which will warn you every time a memory leak occurs. I think it's very handy but it's worth acknowledging that behavior before merging.


vIiRuS:
I noted this as a contribution on your Habitica profile!

PR #304, opened 2015-12-16
Updated Gradle wrapper + build plugins

dlew:
The newer versions of Gradle should run a tad faster, and the latest build plugins are best (especially since newer versions of dataBinder only work w/ the latest Android gradle plugin).

I decided not to tackle the actual app dependencies themselves here since that's a fair bit more involved.


vIiRuS:
I noted this as a contribution on your Habitica profile!

PR #306, opened 2015-12-16
Add travis builds

negue:
we don't have any tests yet, but atleast we can check if the pullrequests can be compiled without errors :)


franzejr:
:+1:

PR #315, opened 2015-12-17
Fixed accidentally introduced potential NPE

dlew:
This can occur when loading an image where one of the layers doesn't load
correctly.


vIiRuS:
Thanks!

I noted this towards your contributor tier.

PR #319, opened 2015-12-18
Compress PNGs

bgiesing:
User ID: 85770757-b546-400f-8ae5-b2cc60880b9a

[Review on Reviewable](https://reviewable.io/reviews/habitrpg/habitrpg-android/319)


dlew:
PNGs are compressed as part of the compilation of an APK...

bgiesing:
@dlew Yeah it should help (as I think the compression I used is higher than aapt). Even if not, it will make the repo smaller and faster to clone which is a benefit.

dlew:
> Even if not, it will make the repo smaller and faster to clone which is a benefit.

Git doesn't work that way; when you clone you get the entire history. That means it will include both the uncompressed *and* compressed versions of the images. It actually increases the repo size.

bgiesing:
@dlew Technically you can if somebody does this: http://stevelorek.com/how-to-shrink-a-git-repository.html

But also, you can download only the latest file revision by cloning with `--depth 1`.

negue:
I think for now we won't compress the PNG, maybe once the app is finished with all features we can start to optimize / improve everything else around the app.

I'm closing this for now.

Issue #183, opened 2015-12-11
Widget!


lemoness:
We're getting lots of request for a widget!

jonyeezs:
+1 be good to have a widget. so i don't have to keep opening up the app just to click on a habit or dailies

goldfndr:
Some (most?) of this functionality is available via [tasker-habitrppg](https://github.com/goldfndr/tasker-habitrpg) (for those with [Tasker](https://play.google.com/store/apps/details?id=net.dinglisch.android.taskerm)).

PR #320, opened 2015-12-18
Enable Widget in Manifest

Vorgone:
The widget was not yet wired up in the manifest file and was unavailable to users.


negue:
Yes, it removed since its the old one of @MagicMicky's App, it isn't working anymore, so at first the Widget needs to be "repaired" or yet better we need to create a new one.

Widget Issue #183 (Crossreference)

Vorgone:
Ah, that makes sense.

Thank you.

PR #322, opened 2015-12-18
Fixed HabiticaApplication.exists()

dlew:
It always returned `false` because it was targeting the wrong file.


vIiRuS:
Thanks!

I noted this towards your next contributor tier

Issue #344, opened 2015-12-22
Currency order differs on mobile app and desktop


Data5tream:
On the website it's Gems, gold and silver and on the app it's gold, silver and gems.

camdroid:
Is anyone currently working on this? I can take care of it if not.

Data5tream:
Probably not. If you want, just fix this and open a PR :wink:

_Don't forget to include your UserID in the description_

PR #365, opened 2016-01-06
Reordered currencies to match web version

camdroid:


camdroid:
Habitica User ID: beb555b6-31ab-4da2-b168-0d02d4c6dda9

camdroid:
#344

vIiRuS:
Sorry that this took so long, there was some internal discussion as to whether which order should be used in the future.

I noted this as a contribution in your Habitica profile

PR #374, opened 2016-01-11
Mark ChatMessages as Seen 

### LINKED ISSUE #266: notification for new party chat messages is not cleared after reading party chat
When you read party chat on the app, and then sync on the website, the notification icon still shows that there are unread messages in party chat (I have confirmed that there definitely are not). 

On the website, reading party chat will clear data for the party from the user's `newMessages` object. Before reading party chat:
````
    "newMessages": {
        "f3de857c-03d3-4042-bf30-d452876661f4": {
            "value": true,
            "name": "my party name goes here"
        },
        "3884eeaa-2d6a-45e8-a279-ada6de9709e1": {
            "value": true,
            "name": "a guild"
        }
    },
````
After reading party chat on the website:
````
    "newMessages": {
        "3884eeaa-2d6a-45e8-a279-ada6de9709e1": {
            "value": true,
            "name": "a guild"
        }
    },
````

negue:
closes #266 

I'm not really happy with this but it works .. any suggestions?


vIiRuS:
This looks similar to how it's done for the iOS app. Which part are you not really happy with? ;)

negue:
all of it :D

Issue #132, opened 2015-12-08
Copy UUID and API Key with one tap


franzejr:
On Account Details, It would be nice if we have an one tap action and copy the UUID/API Token to clipboard.

alexandreabeh:
@Alys Looks like this one can be closed - I have also tested the one tap action on my Android device.

PR #382, opened 2016-01-17
fix 132

yishanhe:
#132,  an accessibility enhancement.
You can now click to copy those tokens to clipboard,  in case you want to use them on your phone.

I wanted to use a long click at the beginning, but it turned out that it may need more work-around. So I sticked to one tap

UserID: 68e57be9-0228-4ce6-8157-801a9d19f7f1
Thanks.


vIiRuS:
I like normal tap better than long press anyway. Great! :tada: 

I noted this as a contribution on your Habitica profile.

PR #389, opened 2016-01-25
Don't animate health bar in party member list

### LINKED ISSUE #388: Health bars animate in party member list
When I scroll through the list of my party's members, their health bars animate back and forth as the views are being recycled. I think this looks kinda confusing (since they aren't gaining or losing any HPs) and they shouldn't animate at all.

Should I work on this? It should be an easy enough task for a first contribution.

patheticpat:
This should fix #388.


vIiRuS:
Thanks!

PR #406, opened 2016-02-05
Fix for broken travis build

Vorgone:
The resource amplitude_app_id is missing and is causing build failures.


vIiRuS:
Ah, yes. Thanks!

Can you tell me your Habitica UserID, so that I can note this as a contribution there?

Vorgone:
18be44b2-acf0-47a8-a7af-1dcde2ff09a2

PR #431, opened 2016-02-19
Translation

Dreknor:
my Habitica User-ID:

String-Translation German


vIiRuS:
Hi, Translations are hanled through transifex and merging them in is done by one of the habitica team members with the transifex client.

PR #438, opened 2016-02-25
Due dailies

toshevski:
I added options in settings to show only due dailies.
my Habitica User-ID: fb863d36-9101-4493-9ada-8358d781ee8a

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -21,6 +21,8 @@
     <string name="pref_reminder_header">Daily Reminder</string>
     <string name="pref_reminder_checkbox">Activate Reminder</string>
     <string name="pref_reminder_picker">Set Reminder Time</string>
+    <string name="pref_dailies_title">Dailies</string>
+    <string name="pref_dailies_show_due">Show only due dailies</string>
 
     <!-- Adding tasks -->
     <string name="task_value">Value</string>
--- a/Habitica/res/xml/preferences_fragment.xml
+++ b/Habitica/res/xml/preferences_fragment.xml
@@ -25,4 +25,13 @@
             android:title="@string/pref_reminder_picker" />
     </PreferenceCategory>
 
+    <PreferenceCategory
+        android:title="@string/pref_dailies_title">
+        <CheckBoxPreference
+            android:key="show_due"
+            android:defaultValue="false"
+            android:title="@string/pref_dailies_show_due"/>
+    </PreferenceCategory>
+
+
 </PreferenceScreen>
--- a/Habitica/src/com/habitrpg/android/habitica/helpers/TagsHelper.java
+++ b/Habitica/src/com/habitrpg/android/habitica/helpers/TagsHelper.java
@@ -1,5 +1,7 @@
 package com.habitrpg.android.habitica.helpers;
 
+import android.util.Log;
+
 import com.magicmicky.habitrpgwrapper.lib.models.tasks.Task;
 
 import java.util.ArrayList;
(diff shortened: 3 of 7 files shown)


vIiRuS:
Thanks! Looking good.
Would you maybe be interested in expanding this PR, to also include the other filter options that are available on the website? That would be "Weak" and "Strong" for habits, "dated" and "done" for todos and additionally "grey" for dailies.
As for interface, the iOS app has the selection for these filters in the same view as the filter selection. And persists the selection as a preference (so similar to how you are storing it now).
img_0506

Having this also available on android would be perfect

toshevski:
I can try to do that, but it will be same like this one with options in settings.
And I have few questions:
1. How do you classify weak and strong habits?
2. How do you classify dated and done?
3. And grey for dailies?

Do you have some functions implemented?

vIiRuS:
Since some of these options would only be used temporarily, having them quickly accessible in the filter sidemenu would be rather important

1. weak habits are all habits with a value <= 0. Strong habits are all with a value > 0
2. dated are all todos, that are not completed and have a duedate set. Done are all completed todos
3. grey for dailies is the inverse of the due dailies

robertpateii:
@toshevski Hi Mirche, I'm interested in helping with this feature. Have you made more progress? Waiting on the code freeze? Shall I fork your version of the repository as a starting point?

toshevski:
@robertpateii I was busy doing some work for my faculty, also I work on one project of my own and I didn't have time for this feature. But please use my code as a starting point if it suits your needs. :)

robertpateii:
Hi @vIiRuS, is this what you want to see for the UI?
image
(obviously "foo" should be "Done". I was just testing the click event.)

Also it seems odd to me to persist these selections as preferences. On the website these filters are stored briefly in memory and destroyed on a page reload. Also this menu is separate from the preferences menu. Can you have two preference menus? I would expect these filters would behave like the tags they're adjacent to and last as long as the app is in memory.

If the app is to behave like the website, then I would Mirche's code as is and add these filters so they behave like the tags in the app.

Does that sound right to you?

vIiRuS:
I had unfortunately not considered that, unlike the iOS app, the filter menu is higher in the hierarchy than the individual lists and would then have to contain all three filter types. This way it looks pretty overwhelming and cramped. Ideally there would be a way to have only the matching options available in every list, so that it's simpler. But that would not fit to the drawer menu currently used for filtering.

At the time I implemented the filtering on iOS, I didn't know the website actually had a preference for settings "Due" as the default view for dailies. I made the selections persistent, because that is probably what 99% of the users would want as a behaviour for dailies. Since we haven't had any complaints about that behaviour so far, I figured we should also implement it this way on android. But if others think carrying over the behaviour from the website, I'd also be open for that.
Currently the implementation from Mirche uses a locally stored variable for the dailies filter preference. It should probably use the set from the website, to have a consistent behaviour there?

robertpateii:
Hey Phillip,

Thanks for checking in on this issue.

I agree that *ideally* "dailies default to due" should sync with the website, but I wouldn't want to add more delay for it. Which it would since I've already resolved the merge conflict with Mirche's code. 

If you agree, here's the PR to merge: #530.

Thanks,

Robert

robertpateii:
never mind, i killed that PR for now. Looks like the new filter actually filters out equipment rewards too. I'll poke through Mirche's code some more when I get the chance.

vIiRuS:
@robertpateii I'm currently doing some refactoring, that will result in the equipment rewards not being affected by the filter, so you can for now assume that that is already the case.

But since the next release is a couple of days away, I would prefer to merge in a finished/complete PR instead of the current version.

Since you are taking over, I'll close this PR.

robertpateii:
Great, thanks. In that case I re-opened the pull request with the merge conflict fixes and Mirche's code.  It should be good to go. #531

PR #446, opened 2016-02-28
Single click Tasks FAB

### LINKED ISSUE #410: UI: Single click '+' to add tasks
The current behavior of the '+' button is to bring up a secondary menu asking the user to select between adding a new 'habit', 'daily', 'todo', or 'reward'. While this is very functional, I think for many users _the expected behavior of this button is to simply add a new task task of the same type as whatever list they are current looking at_. 

I suspect that when users are on the todo tab they almost always choose to add a 'todo' when they press the '+' button, and when they are on the 'habits' tab the majority of the time they are wanting to add a habit, etc. The current behavior makes adding a task of any type a 2 step process instead of a one step process, and the smaller closely spaced menu items increases the likelihood of accidentally selecting a wrong item. This is not a big deal if you are only adding a single item, but for anyone adding multiple items (especially true for new users that are adjusting their lists frequently to match their needs, or anyone using the app as a todo list) this doubles the number of steps they have to take. 

The current behavior seem better suited for use in a combined view, where habits, dailies, and todos would all be shown together. In such a case, the intention of the user be more ambiguous and the menu would be better situated. 

I apologize for re-posting if this has been previous discussed elsewhere. I did make an effort to look around and did not see any previous discussion on the topic.

andrewlord1990:
My Habitica User-ID: fb933ec8-daa3-4a33-9a3c-6abf20a291ae

Fixes #410

Currently the tasks page FAB presents you with a menu, which is best suited for a page that displays multiple types of data. However, this means that whichever type of data you are viewing (Habits, Todos etc), you have to press the FAB and then select the right option in the list of small FABs.

This change instead means you simply press the FAB and it goes to the add page for the type of data you are currently viewing. I.e. when on the Habits tab pressing the FAB goes to the Add Habit page.
- The app wouldn't build without updating the Material Dialogs dependency, the build error was due to the previous version not being found. This was the same in my other active Pull Request.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.java
@@ -229,4 +229,8 @@ public class TaskRecyclerViewFragment extends BaseFragment implements View.OnCli
     public String getDisplayedClassName() {
         return this.classType + super.getDisplayedClassName();
     }
+
+    String getClassName() {
+        return this.classType;
+    }
 }
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TasksFragment.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TasksFragment.java
@@ -153,12 +153,21 @@ public class TasksFragment extends BaseMainFragment implements OnCheckedChangeLi
         todo_fab.setOnClickListener(v1 -> openNewTaskActivity("todo"));
         FloatingActionButton reward_fab = (FloatingActionButton) floatingMenu.findViewById(R.id.fab_new_reward);
         reward_fab.setOnClickListener(v1 -> openNewTaskActivity("reward"));
+        floatingMenu.setOnMenuButtonLongClickListener(this::onFloatingMenuLongClicked);
 
         this.activity.unlockDrawer(GravityCompat.END);
 
         loadTaskLists();
 
         return v;
+    }
+
+    private boolean onFloatingMenuLongClicked(View view) {
+        int currentType = viewPager.getCurrentItem();
+        TaskRecyclerViewFragment currentFragment = ViewFragmentsDictionary.get(currentType);
+        String className = currentFragment.getClassName();
+        openNewTaskActivity(className);
+        return true;
     }
 
     @Override


ajfclark:
Sounds like a great change to me.

andrewlord1990:
@vIiRuS: Is there any questions about these changes that I can answer?

veeeeeee:
Hi @andrewlord1990,

Appreciate your hard work on this issue!

The current behavior for the FAB button is in accordance with Google's Material Design guidelines, as we're working on getting Habitica featured in the Google Play store once we've completed building out a few more things.

However, the team has discussed this extensively, and we understand the preference for the tap to show the form associated with the page that is currently being viewed. Instead of the proposed changes here, we think having the FAB open up to the active page on long press is a better option that also helps solve some of the issues people are having with it.

We will definitely reevaluate the user flow analytics for this once the app is feature-equivalent to the website and we're confident it can be a fully standalone experience.

andrewlord1990:
Hi @veeeeeee, thanks for replying to me.

However, where in the Material Design spec does it state that the current behaviour is in accordance, but having the FAB open the form for the active tab isn't? I am very familiar with the Material Design spec.

For example: https://www.google.com/design/spec/components/buttons-floating-action-button.html#buttons-floating-action-button-behavior

Under "Lateral screens", the video on the right shows a screen, where each TAB has it's own FAB. This change works like that, each tab has a corresponding FAB (to open the correct form).

I understand if you just don't want this change, its a shame if you don't want it, but I understand. It seemed there was a lot of interest for this change on GitHub. Also when I spoke to some people I know that use the app and told them I had made this change they were so happy as it is something they had always wanted.

Let me know what you think.

veeeeeee:
Hi @andrewlord1990, we'll look for the source in the documentation, but we're also in direct contact with representatives at Google Play. We can bring up your suggestion when we speak with them next. We are definitely aware of the github discussion around the issue--we're just not seeing that kind of interest/confusion represented in analytics of current user flows.

In the interim, we're happy to merge your PR if it were changed to be a long press rather than a single tap. Not sure that was communicated clearly before--sorry about that.

andrewlord1990:
Hi @veeeeeee.

That makes sense, I can understand wanting to follow Google suggestions! That is interesting about the user flows though, I guess there are people who want it to work how it does now and others who would rather it would work how I have made it work here.

Thanks very much, I will update the PR so it does this just on long press. I will put back the floating menu for regular press.

robertpateii:
Hi, 

I'm a first time user since habitica was habitRPG. I too am frustrated by #410. I can see the logic of the current implementation though. The button has a shadow and floats above the tabs so yeah it should have the same behavior on all tabs. But it's also frustrating because I expect a single tap add. So maybe a wide button slotted in above the top of each list would be appropriate and not violate the current add button?

I don't like increasing clutter, but adding tasks in habitica should be as painless as in other to-do apps and right now it's not. 

Current flow: 
1. Click plus
2. Read all 4 options to find the todo option
3. Try to click the todo option and not the adjacent options
4. Fill out todo
5. Tap save

Competitor flow A:
1. Tap a sibling todo
2. Press enter
3. Fill out todo
4. Press enter

Competitor flow B:
1. Tap plus
2. Fill out todo
3. Tap done

At least specific icons would speed up step #2 as seen in Google Calendar.  And more spacing would speed up step #3 (also seen in Calendar). Habitica crams in 5 options in the same space Google Cal offers 3. 

Long press to add a todo (or whatever current tab is) directly is a good idea but I would never think to try that in any app. 

If there's interest I could put together a little mock-up. I'm not a designer by trade but it's better than this wall of text. 

My user ID is 992e0e95-529d-464f-825f-cb7ea759fa42

Thanks,
Robert
p.s. metric wise you could measure time from tapping plus icon to the next screen loading. It should be split second otherwise users are struggling to find the button they want like me.  Another good metric would be how many users load an add screen, cancel the add and open a different add screen. That would approximate misclicks especially if you limit to non-brand-new users who may just be exploring.

andrewlord1990:
Reimplemented this from scratch, but wanted it in the same PR to see the history of the conversation to explain.

andrewlord1990:
I have removed all the changes I did previously. As I was previously removing the menu functionality for this page, I was removing the menu entirely. Whereas the change now is on long press. This resulted in a very different and also a very simple implementation.

andrewlord1990:
@veeeeeee and @vIiRuS: This task has now been reimplemented as a long click. The implementation was done from scratch as it was very simple (due to leaving the current functionality there for regular click).

vIiRuS:
Sorry this took forever. Thanks for the changes.

Issue #440, opened 2016-02-26
Display quest progress


vIiRuS:
Either health bar or collection progress depending on active quest

saranlert:
Hi, may I take this issue?

vIiRuS:
Sure! I will merge in your other PR tomorrow (saturday). So you could either wait till then or base your work on this on the PR, to prevent merge conflicts :)

IceManX2X:
I'm currently doing the Golden Knight Part 1 with my group but for the last 2-3 days I have not been listed as contributing. It is like my daily Chron is stuck. My habits reset at 11pm each night but it doesn't show me finding any testimonials. I've finished my dailies plus much more so the amount I should've contributed is a lot. Please help.
User API is e01dcb6b-6239-40d1-a6e1-f3606d1109ef
Group ID: b2dd018a-078c-4837-9bf1-af48dde0a52a

Alys:
@IceManX2X I'm seeing from the quest progress messages that you found quest items about 3.5 hours ago at 1458280803585 (epoch milliseconds) or 2016-03-18 16:00:03 in my timezone (GMT+10). 

There would have been none yesterday because your account was affected by a bug that removed quest status information (fixed about 12 hours ago, which is why you've started doing quest progress again).

I'm not sure what the explanation would be for the 1 to 2 days before then and the party's quest progress messages don't go back that far.

If you notice another day of not making quest progress, please comment on the issue at https://github.com/HabitRPG/habitrpg/issues/4252

PR #451, opened 2016-03-06
add quest progress

saranlert:
- hp bar and rage bar for boss
- item list for collection quest
fixing #440


vIiRuS:
Thanks!

PR #466, opened 2016-03-17
add attributes to task creation if auto allocated

schrockblock:
This PR consists of one commit that adds a spinner for setting a task's attribute if task-based point allocation is set for the user. This is a feature I use a lot, and I got tired of having to enter it in on the web. I will submit an iOS PR for the same in the next few days, to maintain parity between the apps.

It also includes the translation strings for the attribute values, either translated directly by me for languages I know reasonably well (en, de, fr), or taken from another Habitica repo with available translations (es, it, pl, pt, zh). I neither speak Lithuanian, nor could I find the translations for attributes in any of Habitica's repos, so that was not included in this PR.

My Habitica User-ID: 19095cf1-7620-4458-95cd-b4bd08fbe306

### DIFF
--- a/Habitica/src/com/habitrpg/android/habitica/ui/activities/TaskFormActivity.java
+++ b/Habitica/src/com/habitrpg/android/habitica/ui/activities/TaskFormActivity.java
@@ -10,6 +10,7 @@ import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.support.v7.widget.helper.ItemTouchHelper;
+import android.text.TextUtils;
 import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuItem;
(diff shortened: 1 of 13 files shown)


Habitica/src/com/habitrpg/android/habitica/ui/activities/TaskFormActivity.java
```diff
@@ -234,6 +248,13 @@ public void onClick(DialogInterface dialog, int which) {
             createTagsCheckBoxes();
         }
 
+        if (HabiticaApplication.User != null){
+            String allocationMode = HabiticaApplication.User.getPreferences().getAllocationMode();
+            if (TextUtils.isEmpty(allocationMode) || !allocationMode.equals("taskbased")){
+                attributeWrapper.setVisibility(View.GONE);
+            }
+        }
+
```
vIiRuS: The allocation mode should rather be passed through the intent, like the other task data.

schrockblock:
@vIiRuS updated the allocation mode logic to be added to the Intent's extras. (Also included a non-breaking change to use `static final` string keys for adding extras to `TaskFormActivity`)

vIiRuS:
Thanks!

I noted this as a contribution in Habitica

PR #467, opened 2016-03-18
Trad. and Simp. Chinese Translate

sujunmin:
Trad. and Simp. Chinese is different.
Please spearate them to two locales, thanks.

My habitica ID: ac6fbe37-b0dc-40d8-ba14-77dde66fbfa8


vIiRuS:
Hi,

translations are handled through transifex. There currently is one translation with the language code zh (called "Chinese"), that is included. Would this be traditional chinese? There is also a Chinese simplified translation, that is not yet complete (with zh-Hans as code).
https://www.transifex.com/habitrpg/habitica-android/dashboard/

sujunmin:
Ok, I will change it in transifex for Chinese (Taiwan), please help me to verify it, thanks.

PR #483, opened 2016-03-25
add attribute to new tasks if auto allocated

schrockblock:
This PR allows selecting an attribute on **new** tasks (as opposed to just existing tasks) if the user has enabled auto allocation. That it was not included in the first PR regarding task attributes was an oversight.


vIiRuS:
:+1:

Issue #480, opened 2016-03-23
Use date pickers for setting due dates


rsquared226:
This is my first GitHub issue that I've submitted, so if I make a mistake, please correct me. :)

Instead of using the current way of picking dates, the app should use the solution that Google has already made for us: [Date Pickers](https://www.google.com/design/spec/components/pickers.html#pickers-date-pickers).

It'll make dates much easier to pick, and you can see which day of the week you choose as well.

jeubank12:
The class type of the Date Picker is already DatePicker.  Do you mean you think that it should be a date picker where you click it and it opens another view instead of an inline spinner?

rsquared226:
Oh, didn't know that.
Yea, in my opinion it would look better if it opened another view.

On Thu, Mar 24, 2016, 8:02 PM jeubank12 <notifications@github.com> wrote:

> The class type of the Date Picker is already DatePicker. Do you mean you
> think that it should be a date picker where you click it and it opens
> another view instead of an inline spinner?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/HabitRPG/habitrpg-android/issues/480#issuecomment-201078773>
>
-- 

Sincerely,
Rahul Ramkumar

Issue #147, opened 2015-12-08
usability: "Every X Days" and "Start Date" can be changed accidentally while scrolling down the screen


Alys:
Not a but, but a usability problem, although not an urgent one.

If you're in a task edit screen, and you scroll down, if your finger happens to go over the Start Date fields, you can change them without realising you've done so.

saraolson:
Adding this to 1.5 as we're redoing the task creation screen then-- so hopefully we can fix this inherently with those changes.

PR #486, opened 2016-03-26
Enhancement datepickerdialog

jeubank12:
solution to #480 and inadvertently #147 

Fixes start date and due date to use DatePickerDialog instead of the spinners.  Users select the EditText for a popup dialog.  Able to choose "Today" or any date with the default android picker.  String "Today" is translated into existing languages.  Without spinners, the scrolling issue mentioned before is also removed

Also my Android Studio made some formatting changes to the code on its own, apologies for that.

my Habitica User-ID: kskid19


jeubank12:
If its better for me to do no-ff just let me know how I can use git better! :D

vIiRuS:
No need for any git changes. Thanks!

jeubank12:
fcb80135-1ac0-458c-97d5-d3a3e53b11ea

PR #489, opened 2016-03-30
allow reordering of tasks

schrockblock:
This PR allows reordering of tasks by long pressing and dragging. It uses the same API call as the website. A matching PR for iOS will be submitted over the weekend.

My Habitica User-ID:
19095cf1-7620-4458-95cd-b4bd08fbe306

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/tasks/HabitItemRecyclerViewAdapter.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/tasks/HabitItemRecyclerViewAdapter.java
@@ -8,14 +8,8 @@ import android.graphics.Rect;
 import android.support.annotation.NonNull;
 import android.support.v4.content.ContextCompat;
 import android.support.v7.widget.RecyclerView;
-import android.text.Spannable;
-import android.text.SpannableStringBuilder;
-import android.text.Spanned;
-import android.text.method.LinkMovementMethod;
-import android.text.style.URLSpan;
 import android.view.Gravity;
 import android.view.LayoutInflater;
-import android.view.MotionEvent;
 import android.view.TouchDelegate;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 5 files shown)


vIiRuS:
Awesome work so far! However, there are still some issues:

1. the batch-update route will be deprecated soon. Could you change the code to use the /user/tasks/{id}/sort route directly instead?
2. there currently is no feedback for when the dragging has started. Can the cell grow a bit when the dragging starts, so that it appears like it is being lifted up?
3. For some reason a lot of the moves I did, did not stick. The request looks be correct, but the response still has the task in the old position and on the server it doesn't really seem to be changed.

schrockblock:
1. Delighted to! Can you point me to the documentation for this, or give me an example request? So far I've just been looking at what requests the site makes, and I can't find an example of the sort endpoint.

2. What device are you using to test? I get haptic feedback when I long press on my Nexus 5X. This implementation uses Google's default for rearranging items in a RecyclerView, which doesn't give you too much control, but there are some things we can do. Certainly we can do something similar to what's in the checklist RecyclerView. Not sure it's necessary, since Android users will be used to the current functionality, but if you and the Habitica team designers think it'd help, I'm happy to implement it :)

3. Agreed, in my past few days of using this, I've had the same problems. Planning to do some more in depth debugging on it shortly.

vIiRuS:
1. You can find the api doc here: https://habitica.com/static/api#!/user/user_tasks_id_sort
2. Motorola Moto e. It didn't give me any haptic feedback, so I'm assuming not all devices are capable of doing that? In that case, having a visual cue would help a lot.

schrockblock:
Great. I'll take a look at the docs, and get some visual feedback going. Will ping you when the PR is ready for review.

Alys:
> For some reason a lot of the moves I did, did not stick. The request looks be correct, but the response still has the task in the old position and on the server it doesn't really seem to be changed.

This might not be relevant, but on the website, sorting will often fail if there is a filter in place (e.g., viewing To-Dos for only a particular tag, or only dated To-Dos). It's because the website's filtered view of the To-Dos has different array indexes than the actual array of To-Dos that the server knows about. Sort of related: https://github.com/HabitRPG/habitrpg/issues/6839#issuecomment-193985706

I don't know if that's at all relevant on the app though.

jleaders:
Is this available in the google play store now? I have a galaxy S6 and the habitrpg app and would be happy to use this feature. If not I may be able to help test or something if you need as well

vIiRuS:
No, this feature isn't merged into the app yet.

@schrockblock do you need any assistance on this?

schrockblock:
Don't worry! This feature is coming. I just have had some other pressing projects come up. I expect to get this done before mid June.

vIiRuS:
Hey @schrockblock 
Do you maybe need help on this? If you don't have the time to complete this PR, I could take it over and finish it. You would of course still get credit for the partial solve.

schrockblock:
I'm working on it today and tomorrow, and that should do it. Going to close this PR since it's so old now and has too many conflicts, and will reopen another one when I've got it working. Thanks @vIiRuS!

PR #515, opened 2016-04-22
chore: add Habitica connection properties to gradle build

nivl4:
my Habitica User-ID: 51102111-ed4a-4f13-972b-705a3c1e5808

Add Habitica gradle properties file to the gradle build process to
enable debugging against a local instance of Habitica server.


vIiRuS:
Thank you!

Noted this as a contribution in your habitica profile.

PR #518, opened 2016-04-26
Fix to minimum level to change class

Atomk:
I hope I edited the right file and did a good thing.

my Habitica User-ID:
d8fc8d8b-a10f-416a-b089-587db2af9e4d


vIiRuS:
Level 100 is actually correct in this case. What this refers to is the option to change your class at any time for the cost of 3 gems. Subscribing doesn't enable the class system below level 10.

Atomk:
Ops, I'm sorry, But it's not true you can't change class before level 100, and everybody can change class not only subrscribers. Or at least my experience with the website and the Wiki say so...

PR #525, opened 2016-05-07
Recieve the boot command so that we can reschedule alarms.

Jawnnypoo:

Hey there, love the app. So, I have created a few apps that show a daily reminder, and my wife noticed that she was not getting her daily reminders after rebooting her phone. Turns out it is the same issue I have run into many times before, including in [this app](https://github.com/Jawnnypoo/open-meh)

The problem is that all alarms scheduled with AlarmManager get canceled/destroyed when the user reboots their device. Therefore, you must register to receive the on boot event on the device so that you can reschedule the notification. 

This will fix the issue for those individuals who may have reset their devices and not yet opened up the app and given it a chance to reschedule notification delivery. Therefore, we no longer need that check in MainActivity. 

I also went about centralizing the code for scheduling the notification since it was duplicated both in the preference fragment as well as in MainActivity. This way if it ever changes, it will just need to be changed in one place. 

One caveat to all this is that your receiver only gets the boot event if the app is installed on internal storage, as external storage and apps get mounted after the boot command is issued. So, users who have the app installed on external storage will have the issue with their notifications not getting rescheduled. I would suggest restricting the app to only be installed on internal storage. Let me know, and I can add that change to this PR.

### DIFF
--- a/Habitica/AndroidManifest.xml
+++ b/Habitica/AndroidManifest.xml
@@ -16,6 +16,8 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+    <!-- For rescheduling notifications -->
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="com.android.vending.BILLING" />
 
     <application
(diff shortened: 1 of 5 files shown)


vIiRuS:
Hi,

This is looking good so far!

Do you know how other apps deal with the problem with external storage? Since several users explicitly requested the option to install the app on the sd card, I'd prefer to not take that option away again if possible. Is there maybe another workaround for this?

Jawnnypoo:
I would say you have two other options:

1. Add back in the check that was occuring when the app starts. (notifications will get rescheduled as soon as the app is started
2. Register also for some other broadcast event, such as detect when the user plugs in their phone and reschedule then. 

I think option 2 would be better, since it would mean that even if a user did not open the app, their alarms would still be scheduled as soon as they plug in their device. I can make either of these adjustments as needed. Option 2 requires no additional permissions

vIiRuS:
Is there a way to check if the notifications are already scheduled? So that they would either get rescheduled when the app launches or the user plugs in their phone, depending on what happens earlier.

Jawnnypoo:
Yeah, I think that would work. I'll get it updated to have that functionality as soon as I have time

PR #527, opened 2016-05-07
Update to butterknife 8.0.1

Jawnnypoo:

Title says it all. Mostly a find/replace, but Jake did change the API so that you need to store the unbinder object within fragments in order to unbind them.

The dependency also now requires `apt` which I was surprised the app did not already use!

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.application'
 apply plugin: 'com.android.databinding'
 apply plugin: 'io.fabric'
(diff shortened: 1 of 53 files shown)


vIiRuS:
Cool!

So only fragments require the unbinder object to unbind again?

Also it would probably make sense to move the binding and unbinding to the BaseFragment class, so that it doesn't have to be repeated in each fragment (should have done that a while ago).

Jawnnypoo:
Yep, only needed for fragments since they have such a weird lifecycle. 

It has been done!

vIiRuS:
Awesome!

Noted as a contribution

PR #529, opened 2016-05-07
Update guild state

Jawnnypoo:

So, this one fixes an issue where if you were to Join or Leave a guild while viewing the public list, it would not actually update the UI at all. This took a bit of refactoring to get working mostly due to the fact that the ViewHolder was keeping track of things that belonged in the Adapter. Typically, it is best practice to perform API actions within the fragment or activity so that the adapter or view holder just keep up with UI state. In this case, I merely refactored the API updating logic to be within the adapter, with the click listeners defined within the adapter. When a click occurs, the Group associated with the view holder is pulled from the view holders tag and the API operation is perfomed within the adapter. When the operation completes, `notifyItemChanged` is called on the items position so that only one item in the list will update. 

I also went ahead and updated the API call to leave a group so that it was correctly returning Void instead of a Group (as defined in the 204 response code). This was mostly for my own sanity, since I was first attempting to read that returned group when the user left the group and it was resulting in some null pointer exceptions.

I also added an `equals` and `hashcode` methods on the Group model so that I could check a group coming back from the API (different object) against a group in the list. These methods were autogenerated within Android Studio and read just the `id` of the group.

Some of the XML changes are just my autoformatter going nuts, sorry about that.

### DIFF
--- a/Habitica/res/layout/item_public_guild.xml
+++ b/Habitica/res/layout/item_public_guild.xml
@@ -1,44 +1,47 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:orientation="vertical" android:layout_width="match_parent"
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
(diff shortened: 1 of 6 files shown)


vIiRuS:
Thanks!

Noted as a contribution

PR #539, opened 2016-05-23
feat: add GIF supported AvatarView

nivl4:
my Habitica User-ID: 51102111-ed4a-4f13-972b-705a3c1e5808

- change avatar image generation to custom avatar view with layers of `Drawee`
- image cache and memory management provided by default with [Fresco](http://frescolib.org/)
- add support for customizable item position eg.`weapon_special_critical`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ android:
     # Uncomment the lines below if you want to
     # use the latest revision of Android SDK Tools
     # - platform-tools
-    # - tools
+    - tools
 
     # The BuildTools version used by your project
     - build-tools-23.0.3
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -117,6 +117,13 @@ dependencies {
     //Analytics
     compile 'com.amplitude:android-sdk:2.7.1'
 
+    // Fresco Image Management Library
+    compile('com.facebook.fresco:fresco:0.10.0') {
+        exclude module: 'bolts-android'
+    }
+    compile('com.facebook.fresco:animated-gif:0.10.0') {
+        exclude module: 'bolts-android'
+    }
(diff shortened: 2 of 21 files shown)


vIiRuS:
Cool!

Have you done some tests if this negatively impacts performance / memory usage?

We decided against using hungarian notation at some point and want to avoid it going forward (I still have to do a pass through the current code and fix all remaining usages). For this to get merged, could you change this? As for reasoning why, I'll refer to [Jake](http://jakewharton.com/just-say-no-to-hungarian-notation/)

Having both Picasso and Fresco feels weird. It would probably make sense to completely switch to fresco, if that provides more features while also offering everything we use from picasso.

nivl4:
Hungarian notation removed per project coding standard.

In terms of the performance, I didn't notice any negative impacts. As image cache/memory is managed by Fresco library, if there's any performance problems it would probably be fixed in the Fresco project.
I did notice that the memory usage would increase a little bit to cache GIF frames, but would later be released/managed by Fresco. CPU usage would also increase a little bit (less than 1%) for GIF.

Before:
original

After:
new

About migrating everything to Fresco and deprecate Picasso related code, I was hoping to find time to do that in a different pull request once changes in this request has been reviewed and accepted.

nivl4:
@vIiRuS Travis CI build failed because of the new tests added recently (Commit:
a4e7fc2682f9a2e2d2814e6d341c3a48ea0253ce), hence is outside the scope of this pull request.

vIiRuS:
Yes, I also noticed that. I'll be traveling over the weekend, so I'll test and probably merge this in early next week.

I also agree that switching to fresco from picasso isn't necessarily in the scope of this PR. I just wanted to make sure that that was a (good) option.

PR #552, opened 2016-05-29
Added notifications for todos

TheHollidayInn:
my Habitica User-ID:

-I added a Singleton task alarm managers that will listen to created/updated todos and add an alarm
-The manager is started via task creation form activity or boot via the new TaskAlarmBootReceiver
-The alarms have a pending intent that will call the TodoReceiver - which creates a notification for the user
-If the TaskAlarmBootReceiver is called the TaskAlarmManager schedule alarms is called - this helps when a user restart their phone, we need to reschedule the alarms.


Habitica/src/main/java/com/habitrpg/android/habitica/receivers/TodoReceiver.java
```diff
@@ -0,0 +1,56 @@
...
+
+        createNotification();
+    }
+
+    public void createNotification() {
+        Intent intent = new Intent(context, MainActivity.class);
+        PendingIntent pIntent = PendingIntent.getActivity(context, (int) System.currentTimeMillis(), intent, 0);
+
+        Notification noti = new Notification.Builder(context)
+                .setContentTitle("Your todo " + taskTitle + " is due!")
```
vIiRuS: On iOS we just use the task title for the notification. I would also do that here. Alternatively, I think the taskTitle should be styled differently / surrounded by " ", so that it's clear what the task part is.

PR #553, opened 2016-05-30
Added null check and use Integer object type for maxHp

TheHollidayInn:
my Habitica User-ID:

I was getting a crash when loading marty members -
- After checking that the Stats model can have a null maxHP add a null check
- User the Integer return type instead of int primitive to capture the null value


vIiRuS:
fixed in 96ed75ea0ce47c43b67f0c4c613e9fe70d1fbdcf

PR #570, opened 2016-06-23
Inbox

TheHollidayInn:
Add inbox support to the app. 

my Habitica User-ID:

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -56,9 +56,9 @@
     <string name="about.bugreport">Report a Bug</string>
     <string name="about.source_code">Source Code</string>
 
-	<!-- Network Errors -->
-	<string name="network_error_title">Connection Error</string>
-	<string name="network_error_no_network_body">You are not connected to the internet.</string>
+    <!-- Network Errors -->
+    <string name="network_error_title">Connection Error</string>
+    <string name="network_error_no_network_body">You are not connected to the internet.</string>
     <string name="internal_error_api">There seems to be a problem with the server. Try again later.</string>
 
     <string name="authentication_error_title">Authentication Error</string>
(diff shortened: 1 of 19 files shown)


Habitica/res/values/strings.xml
```diff
@@ -347,4 +347,7 @@ To start, which parts of your life do you want to improve?</string>
     <string name="open_in_store">Open in play store</string>
     <string name="change_class_confirmation">Are you sure you want to change your class? This will cost 3 gems.</string>
     <string name="leaderMessage" formatted="false">Message from %1$s</string>
+
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
vIiRuS: this isn't needed, right?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/menu/MainDrawerBuilder.java
```diff
@@ -32,6 +33,7 @@
     // Change the identificationIDs to the position IDs so that its easier to set the selected entry
     public static final int SIDEBAR_TASKS = 0;
     public static final int SIDEBAR_SKILLS = 1;
+    public static final int SIDEBAR_INBOX = 15;
```
vIiRuS: can you change the IDs to be corresponding to their position? (so inbox is 3, tavern 4, etc.)

vIiRuS:
Currently inbox messages are never marked as read. Could you add, so that they are marked as read when the user opens the inbox fragment? We currently don't have a way to have individual messages marked as unread, so marking all as read when the overview fragment is openened is the only way.

Having an indicator that there are unread messages would also be good. The materialdrawer library should have an option to either add such an indicator or create a custom drawer item with an indicator. The indicator should either just be a dot or a a circle with the number of unread messages.

TheHollidayInn:
Updated!

PR #572, opened 2016-07-05
Social notifications

TheHollidayInn:
my Habitica User-ID:

- This uses two services from the FCM (firebase) documentation to get messages and to get the device token.
- The users is registered for push notifications using the pushNotificationManager in the main activity
- The users can modify notifications via the settings.
- One of the FCM services that receives messages pass the messages to the Push Notification Manager and to a factory which builds the notifications.
- If a notification has an action, the notification response is sent to a receiver

Notifications depend on an update to the Push Notification api. I am submitting a PR for that now.


.gitignore
```diff
@@ -54,3 +54,4 @@ fabric.properties
 Habitica/res/values/secret_strings.xml
 habitica.properties
 habitica.resources
+Habitica/google-services.json
```
vIiRuS: Where is this file generated? Can an example/template file be provided for travis?

Habitica/res/values/strings.xml
```diff
@@ -347,4 +358,12 @@ To start, which parts of your life do you want to improve?</string>
...
     <string name="change_class_confirmation">Are you sure you want to change your class? This will cost 3 gems.</string>
     <string name="leaderMessage" formatted="false">Message from %1$s</string>
+
+    <!-- Local notification actions -->
+    <string name="accept_party_invite">ACCEPT_PARTY_INVITE</string>
+    <string name="reject_party_invite">REJECT_PARTY_INVITE</string>
+    <string name="accept_guild_invite">ACCEPT_GUILD_INVITE</string>
+    <string name="reject_guild_invite">REJECT_GUILD_INVITE</string>
+    <string name="accept_quest_invite">ACCEPT_QUEST_INVITE</string>
+    <string name="reject_quest_invite">REJECT_QUEST_INVITE</string>
```
vIiRuS: These should be marked with `translatable="false"`, so that they are not accidentally translated.

Maybe they should also be moved to strings.constants.xml?

Habitica/src/main/java/com/habitrpg/android/habitica/HabitDatabase.java
```diff
@@ -7,5 +7,5 @@
 
     public static final String NAME = "Habitica";
 
-    public static final int VERSION = 21;
+    public static final int VERSION = 22;
```
vIiRuS: I'll merge this after the inbox PR, so it would be 24.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/notifications/PushNotificationManager.java
```diff
@@ -0,0 +1,160 @@
...
+    private Context context;
+    private HabitRPGUser user;
+
+    protected PushNotificationManager(Context context) {
+        HabiticaApplication.getInstance(context).getComponent().inject(this);
+        sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+    }
+
+    public void setUser(HabitRPGUser user) {
+        Log.v("test", "sdffsdf");
```
vIiRuS: delete this?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/LoginActivity.java
```diff
@@ -214,6 +214,54 @@ private void resetLayout() {
...
+        startActivity(i);
+    };
+
+
+	public static void show(final View v) {
+		v.setVisibility(View.VISIBLE);
+	}
+
+	public static void hide(final View v) {
+        v.setVisibility(View.GONE);
```
vIiRuS: Where is this coming from?

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/notifications/HabiticaLocalNotification.java
```diff
@@ -0,0 +1,15 @@
+package com.habitrpg.android.habitica.helpers.notifications;
+
+import android.content.Context;
+
+import java.util.Map;
+
+/**
+ * Created by keithholliday on 6/28/16.
+ */
+public interface HabiticaLocalNotification {
```
vIiRuS: Maybe turn this into an abstract class, so that each notification doesn't have to reimplement certain code.

It could have a method that takes a NotificationCompat.Builder and then then would finish configuring that and then sending it. Each notification implementation would then configure the builder with their specific text and intents, but wouldn't have to each set the sound, icon, etc.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/LoginActivity.java
TheHollidayInn: This was put here when I rebased.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/notifications/HabiticaLocalNotification.java
TheHollidayInn: Completely agree. I didn't want to optimize too early, but after seeing how all the subclasses turned out, I will definitely make it an abstract class.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/LoginActivity.java
TheHollidayInn: Nevermind. I see what happened.

vIiRuS:
The `google-services.json.example` file will have to be copied for travis. you can check the `.travis.yml` file to see how we did that for the other files.

PR #581, opened 2016-07-22
allow habits, dailies, and todos rearranging

schrockblock:
This PR allows reordering of tasks by long pressing and dragging, and highlights the task when doing so. It uses the same API call as the website (v3).

My Habitica User-ID:
19095cf1-7620-4458-95cd-b4bd08fbe306

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -154,6 +154,7 @@ android {
     }
 
     dexOptions{
+        javaMaxHeapSize "4g"
         preDexLibraries = false
     }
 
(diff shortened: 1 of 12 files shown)


vIiRuS:
When I reorder tasks, it always puts the task in the first place (looking at logcat, the url called is always "move/to/0"), no matter where I drop the task. Tested with Android 6.0

schrockblock:
I'm unable to reproduce this behavior on any my devices/simulators. I suspect, therefore, that the issue may be account specific in some way. Can you provide me with a test account on which this issue occurs @vIiRuS?

vIiRuS:
Reordering tasks now works! Thanks

I noted this as a contribution on your habitica profile.

PR #587, opened 2016-08-02
WIP fullprofile: show avatar / stats / attributes / equipment

negue:
To do:
- [x] Open Profile from Memberlist
- [x] List stats 
- [x] List attributes for member profile
- [x] Show avatar with HP/MP/XP Bars
- [x] List equipment
- [x] List attributes for the user
- [x] Send private message from profile
- [x] Translations
- [x] Pets
- [x] Mounts
- [x] Parse markdown for blurb
- [ ] <del>Achievements</del> -> once the API returns achievements with labels

Also in this Pullrequest:
- Loading get contents with the current device language, now the equipment, pets, potions shows the translated labels


Habitica/res/layout/activity_full_profile.xml
```diff
@@ -0,0 +1,343 @@
...
+        <android.support.v7.widget.CardView
+            style="@style/CardView.Default"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content">
+
+            <LinearLayout
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:orientation="vertical"
+                android:padding="5dp">
```
vIiRuS: Each top-level layout in the cardview should use the CardContent style, so that the right padding is used.

Habitica/res/menu/menu_full_profile.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:context=".ui.activities.TaskFormActivity">
+    <item android:id="@+id/private_message"
+        android:title="PM"
```
vIiRuS: This title should be spelled out, since users relying on accessibilty options might need it

Habitica/src/main/java/com/habitrpg/android/habitica/ContentCache.java
```diff
@@ -19,10 +19,17 @@
...
     private ApiService apiService;
+    private String language;
+
 
     public ContentCache(ApiService apiService) {
+        this.apiService = apiService;
+        this.language = "en";
+    }
 
+    public ContentCache(ApiService apiService, String language) {
```
vIiRuS: I feel like this should be refactored a bit. Right now, the APIHelper has knowledge of the language, but to create a contentCache, you then have to take that language and then also store it in the contentCache.
I think it makes more sense, to give the contentCache an instance of the APIHelper and then have a new getContent() method on the APIHelper. That calls the getContent() method on the apiService using the correct language. That way you can request content without having to pass in the language every time.

Habitica/src/main/java/com/habitrpg/android/habitica/APIHelper.java
```diff
@@ -129,6 +132,8 @@ public APIHelper(GsonConverterFactory gsonConverter, HostConfig hostConfig) {
         Crashlytics.getInstance().core.setUserName(this.hostConfig.getUser());
         Amplitude.getInstance().setUserId(this.hostConfig.getUser());
 
+        languageCode = Locale.getDefault().getLanguage();
```
vIiRuS: should this maybe be the language you set habitica to instead? Eventually we also want to have the app itself be able to be in a specified language. So in those cases, the content should also be in that language. So using the configured language and only using the locale as a fallback might make more sense?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
```diff
@@ -0,0 +1,515 @@
...
+                    .setNegativeButton(android.R.string.cancel, (dialogInterface, i) -> {
+
+                        UiUtils.dismissKeyboard(HabiticaApplication.currentActivity);
+                    })
+
+                    .create();
+
+            addMessageDialog.setView(newMessageView);
+
+            addMessageDialog.show();
```
vIiRuS: Can you refactor this dialog out into an extra method in the activity?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
```diff
@@ -0,0 +1,515 @@
...
+        avatarView.setUser(user);
+        avatarWithBars.updateData(user);
+
+        addLevelAttributes(stats, user);
+
+        petCount.setText(CountEntries(user.getItems().getPets())+"");
+        mountCount.setText(CountEntriesBool(user.getItems().getMounts())+"");
+    }
+
+    private int CountEntries(HashMap<String, Integer> hashMap) {
```
vIiRuS: Why is the method name uppercase?

Habitica/res/layout/activity_full_profile.xml
negue: Then all cards don't have any maring outside, everything looks like theres only one background, no seperate cards

vIiRuS: ? The cardviews should keep their style. But the layout inside of the cardview should use the CardContent style. In this case the LinearLayout with the 5dp padding.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
negue: habit of a C# developer :sweat_smile:

Habitica/res/layout/activity_full_profile.xml
negue: I misunderstood you, sorry, I added the CardContent style to the card itself, which wasn't good at all haha, now its better

negue:
@vIiRuS Updated!

PR #595, opened 2016-08-12
Invites qr code

TheHollidayInn:
my Habitica User-ID:

This adds a QR code with the UserID embeded on the party page when a user is not a member and on the user details preference page.

The user can also download the image to share.

Finally, the user can scan other user's qr codes from the invite to party page.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/PartyInviteActivity.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/PartyInviteActivity.java
@@ -16,6 +16,7 @@ import com.raizlabs.android.dbflow.sql.language.Select;
 
 import android.app.Activity;
 import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
 import android.support.design.widget.TabLayout;
 import android.support.v4.app.Fragment;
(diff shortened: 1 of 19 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/PartyInviteActivity.java
```diff
@@ -129,4 +148,42 @@ public CharSequence getPageTitle(int position) {
...
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+
+        IntentResult scanningResult = IntentIntegrator.parseActivityResult(requestCode, resultCode, data);
+
+        if (scanningResult != null) {
+            userIdToInvite = scanningResult.getContents();
+
+            //@TODO: Inject the user
+            this.apiHelper.apiService.getUser()
```
vIiRuS: The user should probably be fetched from the database instead.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/PartyInviteActivity.java
```diff
@@ -129,4 +148,42 @@ public CharSequence getPageTitle(int position) {
...
+        Toast toast = Toast.makeText(getApplicationContext(),
+                userIdToInvite, Toast.LENGTH_SHORT);
+        toast.show();
+
+        Map<String, Object> inviteData = new HashMap<>();
+        List<String> invites = new ArrayList<>();
+        invites.add(userIdToInvite);
+        inviteData.put("uuids", invites);
+
+        this.apiHelper.apiService.inviteToGroup(this.user.getParty().getId(), inviteData)
```
vIiRuS: Instead of directly inviting the user, it should just insert the user id into the list of IDs. so that the user then has the option to scan another ID if they want to.

PR #600, opened 2016-08-18
Delete and edit tags - Issues: 496, 507

jjbillings:
Added the ability to edit and delete tags from the tag-filter drawer.

my Habitica User-ID: d5de2b4c-caa6-4927-97b2-4ad0bda7e85a (Jack Billings)


vIiRuS:
Looking good so far! One issue I found, was that when you activate a tag and then delete it, the tag is still used for filtering and the only way to then deactivate it is by killing the app and launching it again.

jjbillings:
I will take a look at that this morning!
Thanks,
Jack
On Aug 19, 2016 7:04 AM, Phillip Thelen <notifications@github.com> wrote:Looking good so far! One issue I found, was that when you activate a tag and then delete it, the tag is still used for filtering and the only way to then deactivate it is by killing the app and launching it again.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TasksFragment.java
```diff
@@ -474,4 +579,74 @@ public void stopAnimatingRefreshItem() {
...
+                .create();
+        btnDelete.setOnClickListener((View v) -> {showDeleteTagDialog(alert,tag);});
+        alert.setView(editTagDialogView);
+        alert.show();
+
+        alert.getButton(AlertDialog.BUTTON_POSITIVE).setOnClickListener(new View.OnClickListener(){
+
+            @Override
+            public void onClick(View view) {
+                EditText tagEditText = (EditText)editTagDialogView.findViewById(R.id.tagEditText);
```
vIiRuS: Could you move this handling out into a separate method?

jjbillings:
When I was fixing the issue with tags still filtering after they were deleted, I noticed that the checkboxes in the TaskFormActivity did not reflect updated/deleted tags so I fixed that by refreshing the user when a tag is edited or deleted.

vIiRuS:
Awesome! Nice work.

Issue #334, opened 2015-12-21
Better feedback on skills casted


vIiRuS:
p.e. pickpocketing doesn't display how much gold the user got.

SirLoading:
I have added information about gold gained, if anything else needs to be added I can try working on it.

PR #601, opened 2016-08-19
Skill feedback

SirLoading:
my Habitica User-ID: 673c08ae-b5ee-481d-be54-db4ecbee8183

Issue #334   
Hey! I have added information about gained gold, experience and health points to the notification from casting skill.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/callbacks/SkillCallback.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/callbacks/SkillCallback.java
@@ -2,6 +2,7 @@ package com.habitrpg.android.habitica.callbacks;
 
 import com.habitrpg.android.habitica.events.SkillUsedEvent;
 import com.magicmicky.habitrpgwrapper.lib.models.HabitRPGUser;
+import com.magicmicky.habitrpgwrapper.lib.models.Stats;
 import com.magicmicky.habitrpgwrapper.lib.models.Skill;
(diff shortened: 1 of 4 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/callbacks/SkillCallback.java
```diff
@@ -37,6 +37,9 @@ public void call(SkillResponse skillResponse) {
             this.user.setFlags(user.getFlags());
         }
         if (user.getStats() != null) {
+            usedSkill.xp = user.getStats().getExp() - this.user.getStats().getExp();
+            usedSkill.hp = user.getStats().getHp() - this.user.getStats().getHp();
+            usedSkill.gold = user.getStats().getGp() - this.user.getStats().getGp();
```
vIiRuS: I don't think adding the temporary feedback variables to the skills object is the right approach. Why not put them as additional attributes on the SkillUsedEvent object?

Habitica/src/main/java/com/magicmicky/habitrpgwrapper/lib/models/Skill.java
```diff
@@ -20,5 +20,8 @@
     @Column
     public Integer mana, lvl;
 
+    @Column
```
vIiRuS: The @Column annotation is used to define which attributes are persisted in the database. The temporary response values shouldn't be persisted.

PR #610, opened 2016-08-30
Change icon color

jjbillings:
I noticed that the chat_message layout was using white icons on a white background. I switched those to black icons. Here's a quick before/after:

### Before:
habitica_before

### After:
habitica_after

my Habitica User-ID: d5de2b4c-caa6-4927-97b2-4ad0bda7e85a (Jack Billings)


vIiRuS:
Thanks!

I noted this as a contribution on your habitica profile

PR #613, opened 2016-09-05
Feature: Add Language switcher in settings. 

DanielKaparunakis:
my Habitica User-ID: a84f01a4-4c2b-4dbb-9ed0-08f106dda50c


vIiRuS:
where are the `in` and `iw` translations from? They both don't seem to be in our transifex. All languages in the app have to be in transifex, so that the translations can be managed through that.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/LanguageHelper.java
```diff
@@ -0,0 +1,120 @@
...
+                break;
+            case "zh":
+                locale = new Locale("zh");
+                languageCode = "zh";
+                languageAvailable = true;
+                break;
+            case "zh-rTW":
+                locale = new Locale("zh","TW");
+                languageCode = "zh_TW";
+                languageAvailable = true;
```
vIiRuS: Can this switch statement be simplified?
All languages seem to either have a single string for specifying the language or two, to specify a specific variation. having two variables, that are set in the switch statement and then checked would make the switch part a lot simpler, making it easier to add new languages.
The part following the switch statement could probably look something like this?

    if (second != null) {
        locale = new Locale(first, second);
        languageCode = first+"_"+second;
    } else {
        locale = new Locale(first, second);
        languageCode = first
    }

(Variable names are obvious not a good example)

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/LanguageHelper.java
```diff
@@ -0,0 +1,120 @@
...
+                break;
+        }
+    }
+
+    public boolean isLanguageAvailable() {
+        return languageAvailable;
```
vIiRuS: why not check if `locale` is null? Or am I misunderstanding what this variable does?

vIiRuS:
After changing the language in the app, the settings screen doesn't change it's language. This is probably confusing to the user, as it looks like the language selection didn't work. Is there a way to force the settings activity to reload with the new language?

Also, after a language change, the content api should be called with the new language, so that the content is also available in that language.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/LanguageHelper.java
DanielKaparunakis: _where are the in and iw translations from? They both don't seem to be in our transifex. All languages in the app have to be in transifex, so that the translations can be managed through that._

So when I got the translations to work, I tested all of them. I could not get Hebrew & Indonesian to work so I started researching and found people that encountered similar problems. Look at this question at StackOverflow:

http://stackoverflow.com/questions/8393771/android-not-using-finding-my-hebrew-localization#8393844

It looks like Android looks for values-iw & values-in for Hebrew & Indonesian respectively. So I copied and pasted the resources & renamed them. It's worth noting that if you look at the switch statement, the language codes being sent to the site are still `he` & `id`.

_Can this switch statement be simplified?_

If I understand correctly, I think instead of having a large switch block you want some logic to identify what language is coming in so that way in can be reduced. Most languages are simply and have the string as language code & locale, this could be the default case. Then there would specific cases for specific, problematic, languages. I'll work that in.

vIiRuS: Interesting. Didn't know android had those issues with the hebrew and indonesian translations. I'll look into that.


From my example, the switch statement would stay. But instead of each case looking like this:

    case "pt-rBR":
        locale = new Locale("pt","BR");
        languageCode = "pt_BR";
        languageAvailable = true;
        break;

it would be

    case "pt-rBR":
        first = "pt;
        second = "BR";
        break;

with the code I posted above doing the work of setting `locale` and languageCode

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/LanguageHelper.java
DanielKaparunakis: Should locale ever be null? I think it should be declared with a value of _en_ and then changed, if needed, based on the result of the instantiation of the LanguageHelper class. I meant to declare it with a value of _en_ but left it as null. I'll fix that.

DanielKaparunakis:
Okay, I reduced the cases from 18ish to about 5. Most languages will fall into two general cases which are handled by the default block. All the other cases are special language cases. I eliminated the `language available` variable which I think was the variable you were referring to. That variable assisted with the language process at registration but I reworked the logic to use the resource files instead. If you recall, part of the feature request was to send the language of the phone during setup using an `Accept-Language` header.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/preferences/PreferencesFragment.java
```diff
@@ -197,6 +204,36 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
+            Configuration configuration = new Configuration();
+
+            if (android.os.Build.VERSION.SDK_INT <= Build.VERSION_CODES.JELLY_BEAN){
+                configuration.locale = languageHelper.getLocale();
+            } else {
+                configuration.setLocale(languageHelper.getLocale());
+            }
+
+            getActivity().getResources().updateConfiguration(configuration,
+                    getActivity().getResources().getDisplayMetrics());
```
vIiRuS: Can you add that it also reloads the content with the appropriate new language? Otherwise the content will not have the same language and would only change after a while, which would be a strange user experience.

DanielKaparunakis: What content are you referring to? In the latest commit, the app reloads immediately after the language change so everything is displayed in the new language right after the user changes it. I don't see any content that remains in the old language after the user has chosen a new one.

vIiRuS: Does it also reload the content from the /content API? That was the content I was referring to. But maybe I missed that and it's also already reloading that

DanielKaparunakis: Mm, can you tell me where I can find that in the app so I can test it? As of right now, I do not know what the content API is.

vIiRuS: The content api is the REST api call that provides all the content for the app (p.e. gear, pets, mounts. etc.). It can be synced with `apiClient.apiService.getContent()` when you have an apiClient object. Like it is done [here](https://github.com/HabitRPG/habitrpg-android/blob/develop/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/inventory/customization/AvatarOverviewFragment.java#L64)

DanielKaparunakis: Okay @vIiRuS, it was not making that reload call but I have updated that now.

vIiRuS:
Looks good! If you could fix the merge conflicts, I can merge this in!

DanielKaparunakis:
Had some bumps on the road but got it all done.

vIiRuS:
Thanks!

Notes thid as a contribution in your habitica profile.

Issue #618, opened 2016-09-10
"Pets Found" and "Mounts Tamed" counts shown as "#"


vlad-vassilovski:
I'm working on a fix for the following issue:

A user who has no `pets` collection in their `items` JSON will have a "#" displayed as their "Pets Found" count on their profile instead of 0 like in the web app.

The same can be seen for their "Mounts Tamed" count if they have no `mounts` collection.

### Steps to Repro:
Open the profile for a user who has neither `pets` nor `mounts` in their `items` JSON (e.g. a new user who's never had pets or mounts before).

### Expected:
The "Pets Found" and "Mounts Tamed" counts should be displayed as 0, as in the web app.

### Actual:
The counts are each displayed as "#".

### Screenshots:
Here are the pet and mount counts for a new user who's never had pets or mounts before shown incorrectly in the Android app:

![whitefang_androidapp_hashtagsforpetandmountcounts](https://cloud.githubusercontent.com/assets/21161059/18407868/41f41a2e-76eb-11e6-98ee-dbc0759b1c34.png)

And here are that same user's pet and mount counts shown correctly as 0 in the web app:

<img width="923" alt="whitefang_webapp_rightmountandpetcounts" src="https://cloud.githubusercontent.com/assets/21161059/18407870/462e04e2-76eb-11e6-94cf-f7831adab29b.png">

PR #619, opened 2016-09-10
Fix #617, #618: Correct profile's pet/mount counts

### LINKED ISSUE #617: "Mounts Tamed" count calculated incorrectly
I'm working on a fix for the following issue:

The Android app currently doesn't count mounts with a value of `null` while the web app does seem to. For example, if the `mounts` JSON data looked like this:

```
"mounts": {
    "BearCub-White": null,
    "Owl-CottonCandyPink": true
}
```
Then the Android app's "Mounts Tamed" count displayed in a user's profile would show 1 while the web app's "Mounts Tamed" count would show 2.

### Steps to Repro:
Open the profile for a user who has one or more mounts with a value of `null` in their `items` JSON (I may be wrong, but I believe one way the user may generate such mounts for themselves is by releasing their Gen 1 mounts).

### Expected:
The "Mounts Tamed" count should display the number of mounts the user has ever tamed, regardless of their value in the `mounts` JSON (e.g. as I understand, if the user releases their Gen 1 mounts, their "Mounts Tamed" count should not decrease).

### Actual:
The "Mounts Tamed" count will only sum up the number of mounts that have a value of `true` and ignore any mounts with a value of `null` (e.g. if any Gen 1 mounts have been released, then the "Mounts Tamed" count will no longer include them).

### Screenshots:
Here's an example of this issue in the Android app:

clojo_androidapp_wrongmountcount

Compared to the correct mount count for the same user in the web app:

clojo_webapp_rightmountcount

vlad-vassilovski:
my Habitica User-ID: 540ca157-ca25-457e-8e32-9f8a3a8778b9

The profile's mounts tamed count wrongly ignores mounts with null value.

Also, null/nonexistent pet/mount collections have counts showing as "#".

Both counts should instead be shown as 0 for such null collections.

This fix matches the web app, which does the following:
1. Mounts tamed count = sum mount entries with any value (true, null)
2. Nonexistent pet/mount collections have counts showing as 0

Fix #617, #618

### DIFF
--- a/Habitica/res/layout/activity_full_profile.xml
+++ b/Habitica/res/layout/activity_full_profile.xml
@@ -2,6 +2,7 @@
 <ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:fresco="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/fullprofile_scrollview"
     android:layout_width="match_parent"
     android:layout_height="match_parent">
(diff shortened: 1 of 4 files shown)


vIiRuS:
looks good! If you could resolve the merge conflict, I can merge this in! (You can simply push the chagnes to your existing branch and then the PR will update.)

vlad-vassilovski:
@vIiRuS, thanks for taking a look at this PR. 

I've rebased and it should be good to merge in now.

vIiRuS:
Thanks!

PR #620, opened 2016-09-10
Feature: First day of the week

DanielKaparunakis:
my Habitica User-ID: a84f01a4-4c2b-4dbb-9ed0-08f106dda50c


vIiRuS:
Android has a call to get the first day of the week for the current locale. I think this should be used as the default value instead of defaulting to sunday. That way it is already correct for users in locales where monday is the first day. https://developer.android.com/reference/java/util/Calendar.html#getFirstDayOfWeek%28%29

Also, with this setting the order of dailies in the form should probably also be adjusted to the day the user chose instead of always being monday-sunday.

DanielKaparunakis:
@vIiRuS, your suggestions have been implemented.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/FirstDayOfTheWeekHelper.java
```diff
@@ -8,34 +8,39 @@
...
     private int firstDayOfTheWeek;
+    private int dailyTaskFormOffset;
 
-    public FirstDayOfTheWeekHelper(String firstDayOfTheWeekSharedPref) {
+    private FirstDayOfTheWeekHelper(int dailyTaskFormOffset, int firstDayOfTheWeek) {
+        this.dailyTaskFormOffset = dailyTaskFormOffset;
+        this.firstDayOfTheWeek = firstDayOfTheWeek;
+    }
+
+    public static FirstDayOfTheWeekHelper newInstance(String firstDayOfTheWeekSharedPref) {
```
vIiRuS: Why is this used instead of the constructor? Does it have any benefits?

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/FirstDayOfTheWeekHelper.java
```diff
@@ -8,34 +8,39 @@
...
         switch (firstDayOfTheWeekSharedPref){
             case "1":
```
vIiRuS: Just noticing this now, but Why isn't this handled as an integer? Same with the other occurences. Seems Like it would be easier to use/store it as an integer.

DanielKaparunakis: According to [this](stackoverflow.com/questions/11346916), Android uses strings arrays for both entries and entry values and that's all that's supported. Nevertheless, I went ahead and parsed the parameters from SharedPreferences so I could change the case values to the integer constants resulting in better readability.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/FirstDayOfTheWeekHelper.java
DanielKaparunakis: In this specific class, it decouples the logic of the class from the constructor which allows for easier testing. It is worth noting that I could have also achieved this by adding a new method and placed the logic there. This, however, would require two calls, object instantiation & method call, every time the class is used in the app. In general, static factory methods for object creation are worth considering because: One, unlike constructors, they have names, so instead of newInstance you could use very specific names describing the purpose of the constructor. Something that sometimes the method signature cannot achieve on its own. Two, unlike constructors, they are not required to create a new object each time they’re invoked, meaning you could create a static factory that returns the same instance every time whereas when you use new ClassName() you'll always return a new instance. There is a more in-depth explanation of static factory methods in Effective Java, 2nd edition.

vIiRuS:
Looks good now. Thank you!

Issue #631, opened 2016-09-20
Add user ID to profile


vIiRuS:
The profile should display the user id, so that you can then p.e. copy it from there to invite that user.

hcorion:
I'd be interested in working on this.

PR #632, opened 2016-09-28
Userid in profile

hcorion:
my Habitica User-ID: 48794933-9b73-4b05-8efd-f44fde5aedf3
This fixes issue #631 
 If needed I can add screenshots later. If we wanted it to work slightly different, tell and and I'll fix it up.

### DIFF
--- a/Habitica/res/layout/activity_full_profile.xml
+++ b/Habitica/res/layout/activity_full_profile.xml
@@ -154,6 +154,43 @@
             </LinearLayout>
         </android.support.v7.widget.CardView>
 
+        <android.support.v7.widget.CardView
+            style="@style/CardView.Default"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content">
+            <LinearLayout
+                style="@style/CardContent"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:orientation="vertical">
+
+                <TextView
+                    style="@style/CardTitle"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:text="@string/user_id"
+                    android:textStyle="bold" />
+                <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="vertical"
+                    android:gravity="center_horizontal">
+                    <TextView
+                        android:layout_width="wrap_content"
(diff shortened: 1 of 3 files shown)


hcorion:
screenshots!
habitica-uuid-profile
I tried to make it similar to the way the website displays the UUID.

vIiRuS:
Hey,

Could you change it a bit, so that the user ID is below the profile image and has a header with "User ID"? That way it is a little less confusing as to what the string actually is.

hcorion:
@vIiRuS Sure, How do translations work? I assume we'll want the User ID header to be available in different languages as well.

vIiRuS:
All translatable strings are used through androids resource system and if they are in the strings.xml file, they will be translated once I upload the file to transifex. But in this case that shouldn't be necessary, since we should already have a resource string with "User ID".

hcorion:
How does this look? 
habitica-uuid-profile-new
Of course I'd have to add the copy text to the strings.xml. Is it customary to just update the one translation and have the native speakers translate the rest. Or would I use Google Translate (only fluent in English) for all the translations and get the native speakers to change that translation later if it is wrong.

hcorion:
Alright, let's get this thing merged. New screenshot: 
habitica-bug
I added the copy text to strings.xml, it will still need to be translated into other languages. I'm not sure why the CI is failing, but it doesn't seem to be my fault.

hcorion:
The CI build error, shown here: 
```
/home/travis/build/HabitRPG/habitica-android/Habitica/src/main/java/com/habitrpg/android/habitica/receivers/TaskReceiver.java:35: error: cannot find symbol
        Log.e("EXTRAS", extras.toString());
        ^
  symbol:   variable Log
  location: class TaskReceiver
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
13 warnings
 FAILED
:Habitica:compileRetrolambdaDebug
```
appears to be caused by @vIiRuS 's latest commit https://github.com/HabitRPG/habitica-android/commit/95fba1765bc6bdd81a7f4aba94198fd9582ce7fc

vIiRuS:
Jeah, the build error is ok. Looks good now. Thanks!

PR #636, opened 2016-09-30
Set the currently applied filters by default when creating a new task

### LINKED ISSUE #499: Task creation no longer populates default tags from filter
When I would create a todo or other task, the tags from the current filter used to get applied to the task I was creating. Once I discovered that feature, I liked it for a few reasons:

  * Makes creating batches of TODOs with the same tags easier
  * It's more user friendly to always be able to see the task you just created in the list
  * It streamlines the flow of checking related tasks before creating

After the app got better support for configuring tags explicitly with toggle buttons, the tag values are all off by default regardless of the filter state. Could we switch it back to populate the default tags from the filter?

andrewlord1990:
my Habitica User-ID: fb933ec8-daa3-4a33-9a3c-6abf20a291ae

Fixes #499. When creating a task, default the filters to the currently applied ones. This makes it easier to bulk create tasks with a set of filters, but also otherwise you won't be able to see a task after creating it unless you assigned the same filters to it.


vIiRuS:
Thanks!

I noted this as a contribution on your habitica profile.

andrewlord1990:
Thanks.

PR #637, opened 2016-09-30
After creating a new task, switch to the tab of that type.

### LINKED ISSUE #404: Automatically jump to the tab where you have just created a new item.
For example, I was on "Habits" tab, and I was going to create a new TODO. I clicked the fab and selected "New Todo" and filled the form out with several checklist items, and then saved it.

Intuitively, I would like to check for sure this new item is in the "TODOS" tab and looks like what I want it to be. 
And possibly, I may want to check some subtasks immediately after that. Or I need change a little bit after a second thought.

In those cases,  I have to the swipe to the corresponding tab every time. 
Would it be more user-friendly that we just jump to that tab after we have just created a new item in that tab. 

If this is what everybody thinks we should do, I can work on this. :D

andrewlord1990:
my Habitica User-ID: fb933ec8-daa3-4a33-9a3c-6abf20a291ae

Fixes #404. After creating a new task, switch to the tab of that type.


vIiRuS:
Thank you!

Also noted this as a contribution on your habitica profile.

PR #644, opened 2016-10-20
Clean login layout style

caipivara:
my Habitica User-ID: not sure
- It was hard to edit
- The size of the buttons where a little weird
- The logo at the top with weird padding hide fields in small screen sizes

New one:

screen shot 2016-10-20 at 12 44 33 pm


danielgomezrico:
Im not sure but looks like the tests are failing not because of this commit

vIiRuS:
Thank you for your contribution, but to be honest, I personally don't think this is much of an improvement over the current login screen.

1. How do you change to the login screen? The toolbar seems to be gone where you would switch from registering to logging in and back.
2. The branding should definitely still be there. Right now it's not at all obvious which app this actually is
3. Without the lines below the edit texts they look a lot less like editable text fields.
4. wile the current login screen also isn't very good at this aspect, the buttons still have very different sizes.

danielgomezrico:
@vIiRuS I made some updates and end up with:

activities-mainactivity-10212016111633

danielgomezrico:
The main issue I wanted to fix is that scrolling is needed if you want to see bottom link even  if you have a big screen phone, I checked on nexus 5x and OnePlus two.

caipivara:
@vIiRuS what do you think now?

TheHollidayInn:
I think this looks good. Removing the table views is also helpful.

PR #650, opened 2016-11-01
Remove debuggable setup from gralde

caipivara:

The `debug` builds are debuggeable by default and `release` are not.


vIiRuS:
this was acceidentally added to a commit but has since been removed.

PR #655, opened 2016-11-14
Filter drawer fix

anitawoodruff:
Just delay slightly when we fill the tag filter drawer - from onCreate to onActivityCreated.

Android onCreate docs advise this:

"Note that this can be called while the fragment's activity is still in the process of being created. As such, you can not rely on things like the activity's content view hierarchy being initialized at this point. If you want to do work once the activity itself is created, see onActivityCreated(Bundle)."

https://developer.android.com/reference/android/app/Fragment.html#onCreate(android.os.Bundle)

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TasksFragment.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TasksFragment.java
@@ -107,16 +107,6 @@ public class TasksFragment extends BaseMainFragment implements OnCheckedChangeLi
     }
 
     @Override
-    public void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-
-        if (user != null) {
-            tags = user.getTags();
-            fillTagFilterDrawer(tags);
-        }
-    }
-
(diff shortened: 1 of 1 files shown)


anitawoodruff:
The six failing TaskAlarmManagerTest tests on Travis fail for me locally on the 'develop' branch too, so must have been failing before my changes. Is there a tracking bug for fixing the tests or whatever is causing them to fail? (happy to take a look but don't want to step on any toes!)

anitawoodruff:
fwiw I think the TaskAlarmManager tests are failing due to a bug in Robolectric - see https://github.com/robolectric/robolectric/issues/2669. One workaround which makes them pass is to target sdk 23 instead of 21 in the Robolectric config, so the tests don't hit the unimplemented method. However, that would mean losing Lollipop test coverage, which is still the most popular version of Android out there .. Ideally the tests should be run on all versions of Android which have different Habitica implementations for full coverage.

vIiRuS:
Thanks for the changes!

anitawoodruff:
No worries and thanks! 

BTW robolectric merged my pr which fixes the failing Habitica tests, but I don't think they've made a new release yet.

PR #657, opened 2016-11-17
reduced the number of methods

d954mas:
I tried to disable multidex, but it is near impossible. If your know how to reduce more methods plz write.
methods on start 76590
crashlytics proxy: 74280 (2310)
remove okhttp-urlconnection: 74265(15 wtf? maybe some lib use it)
update facebook 70509(3756)
remove hugo 69694(815)
remove support:gridlayout 69449(245)
update fresco:68869(580)
update retrolambda 68303(566)
update dagger 68303(0)
update butterKnife 68074(229)

total:8516

my Habitica User-ID:7a43eb18-134e-4d2f-b23e-809e336403f2


anitawoodruff:
hmmm looks like the leak-canary dependency actually adds ~1800 methods to our debug dexcount right now - despite them saying it should only add "9, or 0 if you use ProGuard." in their FAQs! (https://github.com/square/leakcanary/wiki/FAQ#how-many-methods-does-leakcanary-add). I am tempted to file an issue because that's plain misleading. The bulk of the methods are from org.eclipse.mat, about 1500 of them, and I can tell it's leak canary adding them because they go away if you remove leak canary (and do a clean build).

I tried upgrading to the latest leak canary 1.5, which removes the org.eclipse.mat methods, but brings in ~1100 new com.square.haha methods and 100 more in the main lib to make up for it.

So, I guess the question is, which is valued more, avoiding multidex in debug builds or having leak canary turned on in debug builds?
with_leak_canary
without_leak_canary
with_leak_canary_1 5_haha

anitawoodruff:
(and this isn't a silver bullet, it still only brings us down to 66163 methods, see 2nd screenshot at the bottom left)

d954mas:
Leakcanary musthave in debug. I think facebook sdk, not needed for debug builds. It brings a lot of methods and not work in debug(bad api key). But it is hard to remove it, because it used in loginActivity layout xml.

d954mas:
Update leakcanary 67874 methods

d954mas:
@vIiRuS why project have picasso and facebook fresko. They are doing same work.

vIiRuS:
We originally used picasso for the images, but switched to fresco for the avatar, because that supports animated layers. And then didn't have the time yet, to completely switch to fresco and remove picasso

d954mas:
ok, i will  try to remove picasso

d954mas:
remove picasso  67250

Habitica/src/main/java/com/habitrpg/android/habitica/HabiticaBaseApplication.java
```diff
@@ -201,17 +198,17 @@ private void setupFacebookSdk() {
     }
 
     private void setupCrashlytics() {
-        Crashlytics crashlytics = new Crashlytics.Builder()
+     /*   Crashlytics crashlytics = new Crashlytics.Builder()
```
vIiRuS: If this isn't used anymore, it should be commented out. If someone needs to reference it, it still exists in the git history.

Habitica/src/main/java/com/habitrpg/android/habitica/userpicture/UserPicture.java
```diff
@@ -1,29 +1,8 @@
...
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
-
 public class UserPicture {
 
-    final protected static char[] hexArray = "0123456789ABCDEF".toCharArray();
+   /* final protected static char[] hexArray = "0123456789ABCDEF".toCharArray();
```
vIiRuS: If we aren't using the class anymore, it should probably be removed completely

d954mas: Ok, i will fixed it. Should i made a new commit for fixes, or squash commit to last

vIiRuS:
Looks good now. Thanks!

PR #661, opened 2016-11-18
Added strings to for correct display in the preferences_fragment; cha…

AlbRoehm:
…nged strings.xml in de and eng; Changed some upper case letters in the strings.tutorial.xml


vIiRuS:
Hi,

Thanks for the PR. We manage the translations through transifex, so manual changes to the translated files can't be done, because they would be overwritten when we sync the translation files the next time. You can find more info on that here: http://habitica.wikia.com/wiki/Guidance_for_Linguists

Can you remove the changes from the translation file and replicate them if needed on transifex?

AlbRoehm:
Sure, the easiest way is to make a new PR ? I'm kinda new to Github pulling ..

AlbRoehm:
Removed the translations. I'll have a look into transifex and see if i can update the strings properly.

vIiRuS:
Thanks!

Noted this as a contribution in your habitica profile.

PR #662, opened 2016-11-19
Todo widget

negue:
Features:
- [x] Todo Widget
- [x] Markdown & Emojis (in labels)
- [x] Tap on title to open the app

I tried to use the current widget, add a simple settings-activity and let it be, but there wasn't any update event (intent) after the selection of what task type you want to see, so I had to split up:
- DailiesWidgetProvider
- DailiesWidgetService
- Templates


negue:
todo_widget

Seems that emoji's not working in my emulator, but they do on my phone 🤔

Habitica/res/layout/widget_todo_list_row.xml
```diff
@@ -0,0 +1,42 @@
...
+
+            <ImageView
+                android:layout_width="@dimen/checkbox_compact_size"
+                android:layout_height="@dimen/checkbox_compact_size"
+                android:layout_gravity="center"
+                android:src="@drawable/todo_unchecked" />
+        </FrameLayout>
+
+        <TextView
+            android:id="@+id/dailies_text"
```
vIiRuS: "todo_text" or "task_text"

Alys: "todo_text" would be my preference for consistency with the wiki and Socialite posts, and some of the website's code. We use "task" as a generic term for Habits and Dailies and To-Dos (also even Rewards in some programming contexts).

PR #663, opened 2016-11-20
fixed intro description strings

d954mas:
fixed hardcode numbers in intro_1_description
my Habitica User-ID:7a43eb18-134e-4d2f-b23e-809e336403f2


vIiRuS:
We manage habitica translations through transifex. I can't merge in these changes, because the next time we sync the translations with habitica, the changes would be lost. You can find out more about this here: http://habitica.wikia.com/wiki/Guidance_for_Linguists

PR #667, opened 2016-12-01
Login incentives

TheHollidayInn:
This PR adds two items:

1. It refactors the the api to use a HabitResponse which allows us to use notification data
2. It adds login notification dialogs. 

screenshot_1480608798
screenshot_1480610095


vIiRuS:
Right now, the app can accidentally display multiple notifications at once. P.e. if two requests are made right after each other and both contain a login incentive notification.
I think a better solution would be to have something like a "NotificationManager" class, that you can add notifications to and then does deduplication and handles displaying different notifications after each other. The code to display/configure the notification views, should then also not be in that manager class, but rather in separate classes. Thoughts?

TheHollidayInn:
I agree with separating the logic to a notification class. I will also add handling for the notification.

Habitica/src/main/java/com/habitrpg/android/habitica/APIHelper.java
```diff
@@ -130,7 +131,8 @@
                     .map(new Func1<HabitResponse, Object>() {
                         @Override public Object call(HabitResponse habitResponse) {
                             if (habitResponse.notifications != null) {
-                                showNotificationDialog(habitResponse.notifications);
+                                PopupNotificationsManager popupNotificationsManager = PopupNotificationsManager.getInstance(APIHelper.this);
```
vIiRuS: Is there a reason for not using dagger ( `@Provides @Singleton` ) here?

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/PopupNotificationsManager.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        HabiticaApplication.currentActivity.runOnUiThread(() -> {
+            if ((HabiticaApplication.currentActivity).isFinishing()) return;
+
+            if (this.seenNotifications == null) {
+                this.seenNotifications = new HashMap<>();
+            }
+
+            for (Notification notification: notifications) {
```
vIiRuS: If there are multiple notifications, this way all would be displayed on top of each other, which would probably look bad. Can we maybe have a simple queueing system, that displays the next notification after the previous one is dismissed until no more notifications are left?

Habitica/src/main/java/com/habitrpg/android/habitica/APIHelper.java
TheHollidayInn: I am just not familiar with dagger, so I could use that?? Let me read the docs.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/PopupNotificationsManager.java
TheHollidayInn: Sure, but that shouldn't be an issue here since a user can only have one login incentive notification at a time and this only create dialogues for login notifications.

PR #668, opened 2016-12-04
Messages when changing class

Pandhariix:
__Pull request to solve issue #524__ 

- When changing class, there's a popup saying: "Are you sure you want to change your class? This will cost 3 gems" 
- After the class change is applied, a popup says "You are now a [CLASS], and you have new default battle gear !"

__my Habitica User-ID: 6ab87be6-4e29-409e-b636-9d074e43ad0a__

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -347,7 +347,7 @@ To start, which parts of your life do you want to improve?</string>
     <string name="opt_out_description">Can\'t be bothered with classes? Want to choose later? Opt out - you\'ll be a warrior with no special abilities. You can read about the class system later on the wiki and enable classes at any time.</string>
     <string name="class_confirmation" formatted="false">Are you sure you want to be a %s?</string>
     <string name="class_changed">You are now a %s !</string>
-    <string name="class_changed_description"> You have a new default Battle Gear that you can change under Equipment !</string>
+    <string name="class_changed_description"> You have new default Battle Gear that you can change under Equipment !</string>
     <string name="choose_class">Choose Class</string>
     <string name="dialog_go_back">Go Back</string>
     <string name="opt_out_confirmation">Are you sure you want to Opt Out?</string>
@@ -371,7 +371,7 @@ To start, which parts of your life do you want to improve?</string>
     <string name="share_raised" formatted="false">I just gained a %1$s %2$s mount in Habitica by completing my real-life tasks!</string>
     <string name="open_in_store">Open in play store</string>
     <string name="change_class_confirmation">Are you sure you want to change your class? This will cost 3 gems.</string>
-    <string name="change_class_equipment_warning">Warning : You will no longer be able to buy equipment from the %s class.</string>
+    <string name="change_class_equipment_warning">Warning: You will no longer be able to buy equipment from the %s class.</string>
     <string name="leaderMessage" formatted="false">Message from %1$s</string>
     <string name="confirm">Confirm</string>
     <string name="market">Market</string>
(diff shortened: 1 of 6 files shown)


Habitica/res/values/strings.xml
```diff
@@ -368,7 +369,7 @@ To start, which parts of your life do you want to improve?</string>
     <string name="share_hatched" formatted="false">I just hatched a %1$s %2$s pet in Habitica by completing my real-life tasks!</string>
     <string name="share_raised" formatted="false">I just gained a %1$s %2$s mount in Habitica by completing my real-life tasks!</string>
     <string name="open_in_store">Open in play store</string>
-    <string name="change_class_confirmation">Are you sure you want to change your class? This will cost 3 gems.</string>
+    <string name="change_class_confirmation">Are you sure you want to change your class? This will cost 3 gems</string>
```
Alys: I think the full stop is best left in. I realise it wasn't there in the suggested text in the issue but we do want it here since it's a full sentence.

Can you please also include the warning from the issue about not being able to buy gear from the previous class?

Habitica/res/values/strings.xml
```diff
@@ -346,6 +346,7 @@ To start, which parts of your life do you want to improve?</string>
     <string name="opt_out_class">Opt Out</string>
     <string name="opt_out_description">Can\'t be bothered with classes? Want to choose later? Opt out - you\'ll be a warrior with no special abilities. You can read about the class system later on the wiki and enable classes at any time.</string>
     <string name="class_confirmation" formatted="false">Are you sure you want to be a %s?</string>
+    <string name="class_changed">You are now a %s, and you have a new default battle gear !</string>
```
Alys: Can you include the full text from the issue? "You are now a [CLASS], and you have new default Battle Gear that you can change under Equipment!"

Pandhariix: Okay ! Though, if you want the full text to be correctly visualised, we might have to display it like that

message

Pandhariix: Would it be okay ?

Habitica/res/values/strings.xml
Pandhariix: Okay ! 
I also added the warning about the gear issue
warning change class

Habitica/res/values/strings.xml
Alys: Yes, that's fine! In fact it's better than having it all in one sentence.

Can you remove the spaces from before the exclamation marks and say "You have new..." rather than "You have a new..."? "Gear" when used like this is a plural noun.

Pandhariix: Okay sure, I didn't understand that in this case it was used as a plural noun.

Habitica/res/values/strings.xml
Pandhariix: Yes !

Habitica/res/values/strings.xml
```diff
@@ -346,8 +346,8 @@ To start, which parts of your life do you want to improve?</string>
     <string name="opt_out_class">Opt Out</string>
     <string name="opt_out_description">Can\'t be bothered with classes? Want to choose later? Opt out - you\'ll be a warrior with no special abilities. You can read about the class system later on the wiki and enable classes at any time.</string>
     <string name="class_confirmation" formatted="false">Are you sure you want to be a %s?</string>
-    <string name="class_changed">You are now a %s !</string>
-    <string name="class_changed_description"> You have new default Battle Gear that you can change under Equipment !</string>
+    <string name="class_changed">You are now a %s!</string>
+    <string name="class_changed_description">You have new default Battle Gear that you can change under Equipment!</string>
     <string name="choose_class">Choose Class</string>
```
Pandhariix: I removed the spaces before the exclamation marks and before the "You"

Pandhariix:
@Alys, should I modify other stuff ?

Alys:
I'm so sorry, I should have replied to this ages ago. I'm happy with the text now and have no further changes for it.

vIiRuS:
Thanks!

Issue #569, opened 2016-06-23
Add icons to indicate tags and reminders to task list


vIiRuS:
If a task has a tag assigned or an active reminder it should have an icon displayed in the task list indicating this.

jonyeezs:
+1

AlbRoehm:
I'm pretty much done with implementing the label icon, but i got quite confused when i realized that the tags field is normally not set when the tasks are loaded. Do i get something wrong was this just not necessary before?

PR #676, opened 2016-12-19
Add tag glyph for tasks

AlbRoehm:

Feature: #569  Added glyphTag and clickListener for the glyph.

glyphtag

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/viewHolders/tasks/BaseTaskViewHolder.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/viewHolders/tasks/BaseTaskViewHolder.java
@@ -3,6 +3,7 @@ package com.habitrpg.android.habitica.ui.viewHolders.tasks;
 import com.habitrpg.android.habitica.R;
 import com.habitrpg.android.habitica.events.TaskTappedEvent;
 import com.habitrpg.android.habitica.ui.helpers.MarkdownParser;
+import com.magicmicky.habitrpgwrapper.lib.models.HabitRPGUser;
 import com.magicmicky.habitrpgwrapper.lib.models.tasks.Task;
 import com.magicmicky.habitrpgwrapper.lib.models.tasks.TaskTag;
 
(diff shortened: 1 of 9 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/viewHolders/tasks/BaseTaskViewHolder.java
```diff
@@ -87,14 +97,26 @@ public void bindHolder(Task newTask, int position) {
...
 
+	@OnClick(R.id.tags_glyph)
+	void displayTagsToast(){
+		String tagString = "";
+		for(TaskTag t:
+		    task.getTags())
+		{
+			tagString = tagString + t.getTag().getName() + ", ";
+		}
+//		Toast.makeText(context, tagString, Toast.LENGTH_LONG).show();
```
TheHollidayInn: Did you mean to leave this here?

AlbRoehm: Yes i meant to leave this. In #569 I explained why it is there. Some how there are no name strings in the tag objects only the tag IDs. So I wanted to wait for a hint on how to get the names properly and not write a workaround or something similar. It is meant to tell the user which tags are on the task.

TheHollidayInn: Ah, I see. Well, this issue probably needs some investigation then. The flow is the following:

1. Tag ids are returned on the task object from the API
2. Retrofit attempts to parse the tag and create a TaskTag model and that model attempts to link with the Task models: https://github.com/HabitRPG/habitica-android/blob/develop/Habitica/src/main/java/com/magicmicky/habitrpgwrapper/lib/models/tasks/TaskTag.java#L51

The Tag models are from Retrofit and the user response. So, maybe the Tags from the user are not being saved correctly. Or, there could be an issue with linking the TaskTag ides to local Tag models. Orrrr. This could be an issue with the deserializer : https://github.com/HabitRPG/habitica-android/tree/develop/Habitica/src/main/java/com/magicmicky/habitrpgwrapper/lib/utils. 

A lot going on there - sorry I don't have more insight.

AlbRoehm:
I thought about comparing the IDs which are in the tag objects to the tag list in the filter somehow to get a connection between IDs and names ... But there should be a more straightforward approach somehow. Anyway I think the tags are quite nice so we can just leave this as a to-do for later or someone with more insight.

TheHollidayInn:
That's what the deaserializer or the TaskModel should do, but yea, that's fine with me. I would just add a todo note above your commented code.

AlbRoehm:
Looks like that something else was implemented ? Too bad you didn't mention that its already fixed ...

Issue #666, opened 2016-11-27
leader's name does not appear in guild and party screens


Alys:
The name of a guild or party leader does not appear on the group's first screen, although if the group has a leader message, you can see the leader's name in the "message from ... " text. 

I suggest that the leader's name be placed under the group's description and above the leader message if there is one, formatted something like this:

**Guild Leader:** TheNameGoesHere

I think there should be a line of whitespace above it to separate it from the guild's description, and also a line of space below it if a leader message exists.

For a party, it should say "Party Leader". 

It should not appear on the Tavern's screen.

saraolson:
This should be fixed whenever we get around to party/guild updates. Guild is planned for 1.6, and that might be a good time to throw parties in too since they're very similar.

emily-keller:
Has this issue been fixed?

saraolson:
@emily-keller this issue hasn't been fixed yet as we never quite got around to implementing the new guilds design

PR #682, opened 2016-12-30
Group leader name

josephnw:
This is my first pull request at GitHub
I hope this may be useful to solve issue #666 
The caption still does not show the type yet (guild or party) and the string value is still only on English (values/string.xml)
my Habitica User-ID: 147dd332-1d66-494f-b1bc-08dd90f62de2


vIiRuS:
Thanks!

josephnw:
Thanks, I hope I may contribute more to Habitica in the future since my life is improved with this

PR #684, opened 2017-01-01
Update robolectric & re-enable alarm manager tests

anitawoodruff:
Previously the AlarmManagerTests were throwing NPEs due to a bug in Robolectric, until @vIiRuS disabled them. 

This change upgrades us to the latest version of Robolectric, fixes things up*, and uncomments the tests since they now all pass.

Ideally I'd like to check all the *ApiTests which also use Robolectric still pass, but I'm not set up to run them locally.. (and I note the Travis build is having npm issues currently? :/)

*From https://github.com/robolectric/robolectric/wiki/3.0-to-3.1-Upgrade-Guide : "RobolectricGradleTestRunner is deprecated and uses should be replaced with RobolectricTestRunner."

Edit: I mistakenly wrote that @TheHollidayInn disabled the tests, but it was actually @vIiRuS , sorry for any confusion!

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -120,7 +120,7 @@ dependencies {
     testCompile "junit:junit:4.10"
     testCompile "org.assertj:assertj-core:1.7.0"
     testCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.5'
-    testCompile "org.robolectric:robolectric:3.2-rc1"
+    testCompile "org.robolectric:robolectric:3.2"
     testCompile 'org.robolectric:shadows-multidex:3.1'
     testCompile "org.robolectric:shadows-support-v4:3.1"
     testCompile "org.mockito:mockito-core:1.10.19"
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -120,7 +120,7 @@ dependencies {
     testCompile "junit:junit:4.10"
     testCompile "org.assertj:assertj-core:1.7.0"
     testCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.5'
-    testCompile "org.robolectric:robolectric:3.1"
+    testCompile "org.robolectric:robolectric:3.2-rc1"
     testCompile 'org.robolectric:shadows-multidex:3.1'
     testCompile "org.robolectric:shadows-support-v4:3.1"
     testCompile "org.mockito:mockito-core:1.10.19"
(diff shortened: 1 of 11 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
```diff
@@ -263,7 +263,6 @@ public void onClick(View view) {
 
         // Load the members achievements now
         apiHelper.apiService.GetMemberAchievements(this.userId)
-                .compose(apiHelper.configureApiCallObserver())
```
vIiRuS: removing this isn't the fix to the compilation problem. all API calls need to be configured, because that also defines that they shouldn't run on the main thread. otherwise there is an error, when the api call is actually used.
I pushed a fix to develop, that fixes the original issue, which is that the GetMemberAchievements API call wasn't wrapped in HabitResponse, which is required since the login incentive PR.

anitawoodruff:
Ah, thanks for explaining, I knew I didn't really know what I was doing
there :)

Great you have develop compiling again. I'll rebase and update this PR
tonight.


On Wed, 4 Jan 2017, 11:49 Phillip Thelen, <notifications@github.com> wrote:

*@vIiRuS* requested changes on this pull request.
------------------------------

In
Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
<https://github.com/HabitRPG/habitica-android/pull/684#pullrequestreview-15097331>
:

> @@ -263,7 +263,6 @@ public void onClick(View view) {

         // Load the members achievements now
         apiHelper.apiService.GetMemberAchievements(this.userId)
-                .compose(apiHelper.configureApiCallObserver())

removing this isn't the fix to the compilation problem. all API calls need
to be configured, because that also defines that they shouldn't run on the
main thread. otherwise there is an error, when the api call is actually
used.
I pushed a fix to develop, that fixes the original issue, which is that the
GetMemberAchievements API call wasn't wrapped in HabitResponse, which is
required since the login incentive PR.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/HabitRPG/habitica-android/pull/684#pullrequestreview-15097331>,
or mute the thread

anitawoodruff:
Rebased and removed the problematic commit.

vIiRuS:
Turns out, 3.2 was properly released yesterday. can you remove the RC1, and then I can merge this in?

anitawoodruff:
Done :) Also, fixed tests running on Fridays :P

vIiRuS:
Thanks!

Issue #686, opened 2017-01-07
Health bar does not restore on level-up straightaway


anitawoodruff:
If health is below 100%, it should be restored to 100% immediately on level up.

However, while the health bar _text_ immediately updates to 50/50 as expected, and the experience bar and graphic also update immediately, **the health bar graphic does not update**. Instead it sticks at its previous level:

![health_not_restored](https://cloud.githubusercontent.com/assets/1918555/21744275/031be07a-d50a-11e6-83e4-332b9d6967ea.png)

It remains like this until the next event that triggers a sync.

PR #687, opened 2017-01-07
Fix health bar not restoring immediately on level up

anitawoodruff:
See issue #686 for description of the problem. Either of these commits are sufficient to fix it but both seem wise.

What was happening was on level up, setLayoutWeight was called twice in quick succession for the health value bar, the first time with the _same value as its existing value_ , which nonetheless started an animation to bring it to that same level in 1250ms, so when setLayoutWeight was called again with the new (restored) value, this was subsequently overridden by the animation that was previously set, bringing it back down to the previous level.

These commits ensure any ongoing animations are cleared when setLayoutWeight is called and also that we don't bother setting an animation if the layout weight is already as desired.


vIiRuS:
Thanks a lot!

For the future: If the issue you are resolving doesn't have a github issue yet, you don't have to create an issue before submitting the PR. if the PR already includes the issue description, that's enough! :)

PR #696, opened 2017-01-15
changed PlayerMinStats.getCleanedClassName() to display correct class strings

AlbRoehm:
before these changes names where set from _class.toString() and not looked up in the strings.xml

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -201,7 +201,7 @@ android {
 
     dexOptions {
         preDexLibraries false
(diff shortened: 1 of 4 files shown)


vIiRuS:
an you rename the method? maybe "getTranslatedClassName"? Just to avoid confusion/future issues if someone uses the method without realizing it will return a translated string if the language is different.

AlbRoehm:
no problem, give me second ...

vIiRuS:
Thanks!

PR #707, opened 2017-01-23
WIP: handle task logic in ChallengeDetailActivity // clean-architecture

negue:
In order to be able to use the task button (habit +/-, todo etc) in the ChallengeDetailActivity, we need to refactor how we fire up those api calls, save the new data or even play the sounds.

Most of the boilerplate is from the clean-architecture branch.

Current changes:
- private ApiService
- renamed ApiHelper to IApiClient
- ApiClient now handles `.compose(configureApiCallObserver())` calls, just returns the result without `HabitResponse`
- extracted multiple UseCases:
   - BuyReward
   - CheckClassSelection
   - ChecklistCheck
   - DailyCheck
   - DisplayItemDrop
   - HabitScore
   - LevelUp
   - NotifyUser
   - TodoCheck


TheHollidayInn:
This looks great!! I'm super excited.

My only concern is for the directory structure of the data. 
- Should we do: data/repositories 
 - Or even: data/tasks
       - /local
       - /remote

Habitica/src/main/java/com/magicmicky/habitrpgwrapper/lib/api/ApiService.java
```diff
@@ -100,6 +100,9 @@
     @GET("tasks/user")
     Observable<HabitResponse<TaskList>> getTasks();
 
+    @GET("tasks/user")
+    Observable<HabitResponse<ArrayList<Task>>> getUserTasks();
```
vIiRuS: Why did you add this call a second time, with a different return type?

Habitica/src/main/java/com/habitrpg/android/habitica/interactors/UseCase.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Builds an {@link rx.Observable} which will be used when executing the current {@link UseCase}.
+     */
+    protected abstract Observable<T> buildUseCaseObservable(Q requestValues);
+
+    @SuppressWarnings("unchecked")
+    public Observable<T> observable(Q requestValues) {
+        return this.buildUseCaseObservable(requestValues)
+                .subscribeOn(postExecutionThread.getScheduler())
```
vIiRuS: this causes all usecases to be run on the main thread (since that's the postexecutionthread we're using). In the clean-architecture branch, I did this because realm objects always have to be handled on the same thread and this was a working temporary solution. but we should probably find a way to at least have some of the usecases run on a background thread. Suggestions for how to implement this best are appreciated.

vIiRuS:
@TheHollidayInn depends on if we want to put other things than the repositories in `data`. if not, I'd probably rename `data` to `repositories`. if we want to add other things, then going with `data/repositories` sounds good. I wouldn't split it further at this point, because I don't think we'll have more than a handfull of different repositories.

TheHollidayInn:
Well, from the code above it looks like for each model we will have a DBFLow, local repositroy, and task repo implementation

vIiRuS:
I wouldn't create a repository per model, but rather group them into sections. so that there might be one "Social" repository and a "Inventory" one. that each handle working with the matching data.
for each one we'll have something like this.

- Repository interface - data/Repository.java
- Repository implementation - data/impl/RepositoryImpl.java
- Local Repository interface - data/local/LocalRepository.java
- Local Repository implementation - data/local/impl/LocalRepositoryImpl.java

so while there are 4 files per repository, there is only 1 file in each folder.

TheHollidayInn:
Hmm, I don't think I would do that. I think Challenges should have their own, Tasks, Users, Groups, etc.

vIiRuS:
Not saying they shouldn't. but we p.e. don't need an "EggsRepository". that could go in an Inventory Repository or something like that. That's what I meant.

TheHollidayInn:
Ah, yes. I agree with that.

Habitica/src/main/java/com/habitrpg/android/habitica/data/implementation/TaskRepositoryImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.ArrayList;
+
+import rx.Observable;
+
+
+public class TaskRepositoryImpl extends BaseRepositoryImpl<TaskLocalRepository> implements TaskRepository {
+
+
+    private APIHelper apiHelper;
```
negue: yeah I guess it would be better to use only the `apiclient` , but I'd like to implement the same interface of `ApiService` which would result that `configureApiCallObserver`  has to return `HabitResponse<T>` again

or we'd have two interfaces one with `HabitResponse<T>` and one without, what do you think? @vIiRuS

Habitica/src/main/java/com/magicmicky/habitrpgwrapper/lib/api/ApiService.java
negue: I guess I thought at first thats easier to use the ArrayList<Task>than a TaskList (which has all Tasks in a HashMap) , gonna revert this again^^

Habitica/src/main/java/com/habitrpg/android/habitica/interactors/UseCase.java
negue: I'm gonna try to get an overview of a simple UseCase and what subscribe/observeOn are called and then decide what to add and what not.

My first suggestion would be at first: let the usecase handle if it wants to use the subscribeOn or not?

negue:
@vIiRuS I added some more use-cases, should something like `EventBus.getDefault().post(new TaskUpdatedEvent(event.Task)); ` be added inside an use-case or stay inside a `.subscribe()`-call?

Habitica/src/main/java/com/habitrpg/android/habitica/interactors/CheckClassSelectionUseCase.java
```diff
@@ -0,0 +1,78 @@
...
+        bundle.putInt("hairBase", user.getPreferences().getHair().getBase());
+        bundle.putString("hairColor", user.getPreferences().getHair().getColor());
+        bundle.putInt("hairMustache", user.getPreferences().getHair().getMustache());
+        bundle.putInt("hairBeard", user.getPreferences().getHair().getBeard());
+        bundle.putBoolean("isInitialSelection", event.isInitialSelection);
+        bundle.putString("currentClass", event.currentClass);
+
+        Intent intent = new Intent(HabiticaApplication.currentActivity, ClassSelectionActivity.class);
+        intent.putExtras(bundle);
+        HabiticaApplication.currentActivity.startActivityForResult(intent, SELECT_CLASS_RESULT);
```
negue: yeah its a bit too much, but the ClassSelection is used in multiple ways so it kinda "screams" to be a usecase 😄

Habitica/src/main/java/com/habitrpg/android/habitica/interactors/ListTasksUseCase.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Inject
+    public ListTasksUseCase(TaskRepository taskRepository, ThreadExecutor threadExecutor, PostExecutionThread postExecutionThread) {
+        super(threadExecutor, postExecutionThread);
+        this.taskRepository = taskRepository;
+    }
+
+    @Override
+    protected Observable<ArrayList<Task>> buildUseCaseObservable(ListTasksUseCase.RequestValues requestValues) {
+        return taskRepository.getTasks(requestValues.taskType);
```
negue: `ListTasksUseCase` was a test UseCase-Class, gonna remove it

PR #710, opened 2017-01-25
Merge multiple system skill cast messages into one message 

AlbRoehm:
I find it rather annoying to see "only" system messages in the group chat, especially if they are multiples of someone casting a spell, so i added a algorithm which looks for multiple skill cast and merges these messages into one message with a trailer of the amount of casts. 
Hope u guys find this as usefull as i do, but i'm for open for changes requests and other ideas...

For example:

Haggy casts Tools of the Trade for the party.
Haggy casts Tools of the Trade for the party.
Haggy casts Tools of the Trade for the party.

-->

Haggy casts Tools of the Trade for the party. (3 times)


vIiRuS:
While I like the idea, could you change the way it's done/presented? Instead of merging multiple messages into one, consecutive messages should be hidden. So that the one that is shown would have something like "Show X additional messages" as a button at the bottom.

AlbRoehm:
I think your idea is also a possible approach, but i don't get which Information a user should get from these additional shown messages, because they'll all show the same information. I would rather discuss hiding systems messages completely for people not interested in that kind of stuff, controlled by some kind of setting.

AlbRoehm:
@vIiRuS Another thing which came to my mind is that the system messages are not translated at all ... Don't know why i didn't see this earlier. So probably some more work, or should this be handled serversidely?

Alys:
@AlbRoehm We have an issue for translating system messages at https://github.com/HabitRPG/habitica/issues/5757 but it's not a simple thing because parties can contain players using different languages.

AlbRoehm:
@Alys yeah I can imagine that it's not simple, but on the other hand it's
quite straightforward. You eather have to store the language on the server
and send messages accordingly or send tokens for the different messages an
let the client handle everything. So from my newbie view it's a design
decision and we should  handle the android client the same way as the
webpage, in comparison to compatible views of a MVC pattern. Does this make
any sense? But I can also think of a lot of historically related issues
which make this complex :).

Am 31.01.2017 10:24 nachm. schrieb "Alys" <notifications@github.com>:

AlbRoehm:
Ok, If I got it right you already decided to "let the client handle a
object" so there is not much else to do, then wait for the other issue
being fixed and change the android client afterwards. So maybe we go back
to the initial topping on how to handle duplicates.

Am 31.01.2017 11:11 nachm. schrieb "Alys" <notifications@github.com>:

jamesjpk123:
I back the idea that:

For example:

Haggy casts Tools of the Trade for the party.
Haggy casts Tools of the Trade for the party.
Haggy casts Tools of the Trade for the party.

-->

Haggy casts Tools of the Trade for the party. (3 times)

PR #714, opened 2017-02-02
add checkstyle and pmd

vIiRuS:
Adds gradle steps + configs to start using [checkstyle](http://checkstyle.sourceforge.net) and [PMD](https://pmd.github.io) in order to improve code quality and consistency. The added configs are still open for debate and might not be final.

Key "Features":

1. variable, field and method names must begin with a lowercase letter and can't be single letter.
1. constants have to be all-uppercase
1. package names can't contain uppercase characters
1. tab indentation is not allowed
1. Annotation location is the lines before the thing they refer to
     ```
    @Annotation
    public String thing
    ```
1. anonymous functions can be 60 lines maximum (debatable number, but enforcing some kind of upper limit sounds good for readability
1. Star imports are not allowed
1. redundant and unused imports are not allowed
1. braces style is as follows:
    ```
    if {
        //something
    } else {
       //something else
    }
    ```
1. braces are always required. also for single line if or switch statements
1. function, if/else, loop blocks can't be empty
1. catch blocks can only be empty if the parameter name is "ignore" or "expected"
1. in comma separated lists there has to be a whitespace after the comma
1. classes have this order:
    1. Class (static) variables. First the public class variables, then protected, then package level (no access modifier), and then private.
    1. Instance variables. First the public class variables, then protected, then package level (no access modifier), and then private.
    1. Constructors
    1. Methods
1. when comparing with a string literal, the literal is on the left side
    `"hello".equals(greeting")`
1. No fall through in switch statements
1. variable names can't hide variables that are defined in the outer scope (with exception of setters)
1. catch can't catch `java.lang.Exception, java.lang.Throwable, java.lang.RuntimeException` directly as that's too broad
1. functions can't throw `java.lang.Exception, java.lang.Throwable, java.lang.RuntimeException` directly as that's too broad
1. variables can't be assigned inside subexpressions
    `String s = Integer.toString(i = 2); // this would not be valid`
1. No magic numbers except for a few exceptions. This throws a whole bunch of errors in our code. So while it would make sense, it's also a lot that would have to be changed
1. each switch statement needs a `default` state. If this doesn't make sense for the usecase, it should assert that (the assert part can't be checked and would have to be done by us)
1. A string literal can't appear more than 2 times within the same file
1. for loops can't be deeper than 2. (maybe useful in terms of performance impacts that deeper nesting could have?
1. accessing a field requires the use of `this`. This might be controversial. I'd like to do it, but we can also go the other way. Mostly I'd like to agree on a usage and stick to that.
1. indentation is 4 spaces. Unfortunately this currently causes some issues with the way android studio wants to indent some rxjava things. So might not be viable if we have to constantly fight the IDE.
1. PMD has checks for [cyclomatic complexity](https://pmd.github.io/pmd-5.5.3/pmd-java/rules/java/codesize.html#CyclomaticComplexity) and [NPath Complexity](https://pmd.github.io/pmd-5.5.3/pmd-java/rules/java/codesize.html#NPathComplexity). Currently they are set with the default values. Using this would require larger refactors for some things.

There are a bunch of other things, but those are more smaller things or checks that are more about performance or code validity (passing a locale to `toUpperCase` p.e.) so those don't really require any discussion, so I didn't list them.

With all these checks and after android studios automatic code import and code formatting we are at around 1200 Checkstyle issues and 750 PMD issues. Although a lot of things are reported in both systems, so the number of issues that would have to be fixed is smaller. Also, most of these things are easily done with android studios built in refactoring tools.

My long term plan for this would be to agree on a config for these tools and then start working on eventually getting the number of issues to 0. Once that is the case, I'd switch them to fail on issues and include them in travis, so that violations are caught quickly. 

Feedback/discussion about any of the mentioned things are probably best done by mentioning the number they have in the list.

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -2,19 +2,26 @@ apply plugin: 'com.android.application'
 apply plugin: 'com.android.databinding'
 apply plugin: 'io.fabric'
 apply plugin: 'com.neenbedankt.android-apt'
-apply plugin: 'com.chaitanyapramod.findbugs-android'
+apply plugin: 'com.noveogroup.android.check'
 
 buildscript {
     repositories {
         mavenLocal()
         mavenCentral()
         maven { url 'https://maven.fabric.io/public' }
+        jcenter()
     }
     dependencies {
         classpath 'io.fabric.tools:gradle:1.21.5'
         classpath 'me.tatarka:gradle-retrolambda:3.3.1'
         classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.6.1'
         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
+        classpath('com.noveogroup.android:check:1.2.3') {
+            exclude module: 'checkstyle'
+            exclude module: 'pmd-java'
+        }
+        classpath 'com.puppycrawl.tools:checkstyle:7.5'
+        classpath 'net.sourceforge.pmd:pmd-java:5.5.3'
     }
 }
 
@@ -276,6 +283,8 @@ tasks.whenTaskAdded { task ->
         task.enabled = false
     }
 }
+
+check { findbugs { skip true } }
 
 apply plugin: 'com.getkeepsafe.dexcount'
 apply plugin: 'com.android.application' //or apply plugin: 'java'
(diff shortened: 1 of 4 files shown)


negue:
LGTM

PR #726, opened 2017-03-08
improve chinese translation

bangbang93:
my Habitica User-ID: 8be41e3f-9c35-4bb5-b846-89878c283f92


vIiRuS:
Hi,

Thanks for the work, but unfortunately this isn't the right channel for this. The translations are handled on transifex, so all translation changes should be made there. You can find more about this process here: http://habitica.wikia.com/wiki/Guidance_for_Linguists

Issue #723, opened 2017-03-05
Class Bonus is incorrectly given for some Battle Gear


Alys:
As reported by Lydy (76220168-3529-4cbe-85a7-caf81e344360) in the Report a Bug guild, a class bonus can sometimes be displayed on the app when the user should not have a class bonus.

I don't know how reproducible this is or under what circumstances it occurs, but Lydy's example is below. With this equipment, there should be no class bonus. Lydy is a Warrior (possibly the [lunarWarriorArmor](http://habitica.wikia.com/wiki/Stop_the_Overshadowing_Stress) quest reward gear is being incorrectly identified as being Warrior class gear?)

```
"items.gear.equipped": {
	"shield": "shield_armoire_mysticLamp",
	"head": "head_armoire_greenFloppyHat",
	"armor": "armor_special_lunarWarriorArmor",
	"eyewear": "eyewear_special_blueTopFrame",
	"weapon": "weapon_special_pageBanner",
	"back": "back_mystery_201702"
}
```

app shows class bonus when there should not be one:
![image](https://cloud.githubusercontent.com/assets/1495809/23584866/de1d6cd0-01b9-11e7-8a79-391ba2b07386.png)

website correctly shows no class bonus:
![image](https://cloud.githubusercontent.com/assets/1495809/23584867/ea543c36-01b9-11e7-8caa-4b6691163e0e.png)

----

There's another issue about class bonuses which is not the same as this one but perhaps could be fixed at the same time if someone wants to work on both together: https://github.com/HabitRPG/habitica-android/issues/713

TheHollidayInn:
So, the Lunar Warrior Armor is not Warrior Gear? On a side note, that seems sort of confusing :P

TheHollidayInn:
It doesn't look like a gear issue. It seems that based on being a class you get some static bonus: https://github.com/HabitRPG/habitica-android/blob/develop/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java#L442.

What is the correct formula for calculating class bonuses?

PR #735, opened 2017-03-29
Fixed class bonus calculation

TheHollidayInn:
my Habitica User-ID:

Fixes: #723

This adds the klass field to gear and checks to see if there is a bonus from the item based on the klass field

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
@@ -424,27 +424,38 @@ public class FullProfileActivity extends BaseActivity {
 
             // Calculate class bonus
             String itemClass = i.getKlass();
+            String itemSpecialClass = i.getSpecialClass();
+
+            if (itemClass.isEmpty() && itemSpecialClass.isEmpty()) {
+                continue;
+            }
+
+            float classBonus = 0.5f;
+            Boolean userClassMatchesGearClass = itemClass.equals(user.getClass());
+            Boolean userClassMatchesGearSpecialClass = itemSpecialClass.equals(user.getClass());
+
+            if (!userClassMatchesGearClass && !userClassMatchesGearSpecialClass) classBonus = 0;
 
             if (itemClass.isEmpty()) {
-                continue;
+                itemClass = itemSpecialClass;
             }
 
             switch (itemClass) {
                 case "rogue":
-                    strClassBonus = str_ * 0.5f;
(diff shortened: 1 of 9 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
```diff
@@ -416,40 +421,39 @@ public void gotGear(List<ItemData> obj, HabitRPGUser user) {
...
+            }
 
-            switch (user.getStats().get_class()) {
-                case rogue:
-                    strClassBonus = strAttributes * 0.5f;
-                    perClassBonus = perAttributes * 0.5f;
+            switch (itemClass) {
+                case "rogue":
+                    strClassBonus = str_ * 0.5f;
+                    perClassBonus = per_ * 0.5f;
```
Alys: I'm probably being blind but I can't see where the user's class is taken into account. For example, `case "rogue"` means that the str and per bonuses are calculated for any gear that belongs to the rogue class (I think), but where does it say that the bonus is applied only when the user's class is rogue?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.java
```diff
@@ -424,27 +424,38 @@ public void gotGear(List<ItemData> obj, HabitRPGUser user) {
...
+            if (itemClass.isEmpty() && itemSpecialClass.isEmpty()) {
                 continue;
             }
 
+            float classBonus = 0.5f;
+            Boolean userClassMatchesGearClass = itemClass.equals(user.getClass());
+            Boolean userClassMatchesGearSpecialClass = itemSpecialClass.equals(user.getClass());
+
+            if (!userClassMatchesGearClass && !userClassMatchesGearSpecialClass) classBonus = 0;
+
```
Alys: I think that now, by the time the code reaches this point, classBonus always has the right value for handling this particular gear piece (i.e., if the gear piece belongs to the rogue class and the player is a rogue, then classBonus is 0.5; for any other combination, classBonus is 0). That means that the switch statement is no longer needed. All you need is:
```
	conClassBonus = con_ * classBonus;
	intClassBonus = int_ * classBonus;
	perClassBonus = per_ * classBonus;
	strClassBonus = str_ * classBonus;
```
It doesn't matter that that code is calculating bonuses for attributes that aren't relevant for the user's class because those attributes are going to be set to 0 and so the class bonus for them will end up as 0.

TheHollidayInn:
Updated!

Issue #730, opened 2017-03-21
Specify instance in Login/Register activity, not in build property file.


jcgruenhage:
It would be nice, if people who run their own instance of habitica for privacy reasons wouldn't have to compile the app themselves, but instead could just set the server they want to connect to on the login page.

Alys:
Just as a note about this:

We had that feature on the original version of the app, a long time ago. Non-technical users were confused by it and would sometimes enter their own website addresses (e.g., a personal blog). I suspect that even if a description of the feature was provided on the login page, it would still cause confusion and misuse because people often don't read descriptions. The app needs to seem simple to use and not confusing for non-technical users otherwise they'll be less likely to use it.

jcgruenhage:
Is there screenshot of that somewhere? Even though you're right about confusing non-technical users, I think hiding it away would be a lot better than having to compile the app, especially since this blocks auto-updates, and is not exactly trivial (harder than hosting your own instance with docker). Would there be a starting point, what needs to be changed?

citrusella:
If we're talking the old HabitRPG app (the one that stopped working with API v2's being disabled), this is what it looks like, I believe.

![screenshot_2017-03-29-14-22-46 1](https://cloud.githubusercontent.com/assets/14182450/24470208/1e7989be-148c-11e7-9e43-ed627996b02b.png)

(Mine was filled in with "http://habitica.com" when I opened the app; I just blanked it out for the picture.

And here's the PR that introduced it (as far as my untrained eye can tell): HabitRPG/habitrpg-mobile#271

jcgruenhage:
Ahh well.. If I do a PR on this, there would be at least 3 clicks, including a dialog that everything beyond this point is unsupported, to change this, not a simple text edit field below the password. This really is to confusing for non-tech users :D

Thanks for sending this :)

Alys:
@jcgruenhage  Would you like to describe what the first step in accessing the field would be? For example, would there be a small link on the current login form?

@citrusella Nice work!!

jcgruenhage:
First of all: WOW, I just went ahead to change up the layout to show how I think the UI Stuff for this should be done, and the new login/register activity is damn impressive.

Now, onto business: I'd use a small text view saying "Advanced Login", like this:

![screenshot_20170330-125812](https://cloud.githubusercontent.com/assets/6861958/24501230/e69092d2-1548-11e7-849d-136df899c6d9.png)

[...]

PR #736, opened 2017-04-01
WIP: Advanced login, where the user can specify a custom API endpoint.

jcgruenhage:
A fix for #730, that is not fully functional yet, but very close. Any thoughts, what I should change?

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/data/implementation/ApiClientImpl.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/data/implementation/ApiClientImpl.java
@@ -127,15 +127,15 @@ import rx.schedulers.Schedulers;
 
 public class ApiClientImpl implements Action1<Throwable>, ApiClient {
     private final GsonConverterFactory gsonConverter;
-    private final HostConfig hostConfig;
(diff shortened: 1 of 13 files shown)


jcgruenhage:
#### Needed Fixes:
 - [ ] Login Activity restarts after clicking on Login, but after restart of the app, the user is already logged in.
 - [ ] Advanced register causes a NPE. The field reported as being null shouldn't be null though.

Habitica/src/main/java/com/habitrpg/android/habitica/data/implementation/ApiClientImpl.java
```diff
@@ -127,15 +127,15 @@
...
+    private HostConfig hostConfig;
+    private Retrofit retrofitAdapter;
     private final PopupNotificationsManager popupNotificationsManager;
 
     private CrashlyticsProxy crashlyticsProxy;
     private Context context;
 
     // I think we don't need the ApiClientImpl anymore we could just use ApiService
-    private final ApiService apiService;
+    private ApiService apiService;
```
jcgruenhage: Making those non-final is surely not what we want, but I couldn't find another way to do this, "thanks" to dependency injection.

PR #737, opened 2017-04-04
Update strings.xml

Atomk:
Fixed a typo. Is this the correct way to do it? Someone reported this typo as an issue on Transifex, did you see it or there's the need for someone to manually check the issues (or @-mention you)?

---

my Habitica User-ID: d8fc8d8b-a10f-416a-b089-587db2af9e4d


vIiRuS:
For issues with the source strings, this is the right place. I also get the issue reports via email, but haven't had the time to look into it. Thanks.

Issue #701, opened 2017-01-21
sound effects stop playing if two are played at once (audio themes)


Alys:
If you take some action that produces a sound effect and then before the sound effect has finished playing you take a second action that plays the **same** sound, then that particular sound effect stops playing for all future actions until the app is restarted.

Steps to reproduce both good behaviour and bad behaviour:

1. Wait until you've received all your drops for the day (not essential but makes it easier to hear sounds without interference from the drop notification sound).
1. Choose an audio theme that has long sounds (not essential but makes testing easier). E.g., LuneFox's Theme.
1. Click + on a Habit. Wait for the sound to start and finish completely.
1. Click + on a Habit again. The sound plays again - this is correct behaviour.
1. Click + on a Habit and wait for the sound to start then IMMEDIATELY (before the sound finishes):
1. Click + on a Habit again. The sound does not play (incorrect behaviour).
1. Click + on a Habit again. The sound still does not play (incorrect behaviour).
1. Click anything that causes a different sound (e.g., complete a Daily). That new sound does play (correct behaviour).
1. Close and reopen the app.
1. Click + on a Habit. The sound does play (correct behaviour).

_Reported by SakuraBlossoms64 (911b2b6e-60a9-4574-a99d-285511b243ee) in the Report a Bug guild; confirmed by me._

Alys:
This is a minor bug because it's not good to click tasks too quickly because sync errors tend to happen more often.

GuyKroizman:
@Alys , Could you confirm this was really fixed.
And also would like you to provide your opinion of #738

Alys:
Confirmed fixed.

PR #738, opened 2017-04-08
use-sound-level-of-media-instead-notification

GuyKroizman:
Using the android app, the volume for music/sound of task completed (when I touch a task checkbox), depends on the volume level of "Notification sound". I would expect the volume of that to depend on the volume of "Music, video, games & other media"

It is very annoying as sometime I check the last tasks in bed just before cron and then the app plays this sound and it wakes my wife :)

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/SoundFile.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/SoundFile.java
@@ -54,7 +54,7 @@ public class SoundFile implements MediaPlayer.OnCompletionListener {
 
         try {
             mp.setDataSource(path);
-            mp.setAudioStreamType(AudioManager.STREAM_NOTIFICATION);
+            mp.setAudioStreamType(AudioManager.STREAM_MUSIC);
             mp.prepare();
 
             playerPrepared = true;


GuyKroizman:
Travis CI build failed. I don't think it is related to the change of this PR.
What should I do?

vIiRuS:
This is what we oroginally had, but received multiple complaints that users
expected it to work the other eay around. After changing it a lot less
people complained about it. So while the current solution might not be
ideal, I don't think this is the solution.

-- 
Phillip Thelen
Sent with Airmail

GuyKroizman:
Looking at the file history, that change only happened once.
And anyway It is clearly a mistake to use STREAM_NOTIFICATION for something that is clearly not a notification. (See [documentation](https://developer.android.com/reference/android/media/AudioManager.html#STREAM_NOTIFICATION))

vIiRuS:
It was originally created with the music type. Then changed once to
notification.
It's also clearly not music, so I don't feel like that's a strong argument.
Are there examples of other apps and how they handle these kind of sounds?

-- 
Phillip Thelen
Sent with Airmail

GuyKroizman:
I Could not find good documentation that shows that.
Since we don't see eye for eye on this I will try to get more commenters here.

Also, Do you remember if this change from STREAM_MUSIC to STREAM_NOTIFICATION really fixed #701 ?

vIiRuS:
That change did not fix that issue. (And wasn't intended to fix it). The
fix for that issue was hhe linked commit.
https://github.com/HabitRPG/habitica-android/commit/a30dc2e12af25eebfccc64e1690daa09ec29c4c5
Are you still experiencing the issue?

Alys:
@GuyKroizman I agree with this comment from @vIiRuS. Judging the success/failure of changes by the number of complaints is a powerful tool.

> "This is what we oroginally had, but received multiple complaints that users expected it to work the other eay around. After changing it a lot less people complained about it. So while the current solution might not be ideal, I don't think this is the solution."

GuyKroizman:
I disagree, It would be a powerful tool if it was recorded or measured but since it is not it is a very bad tool. the information only resides in someone's head and used to deflect real issues without any proof.
In contrast to the discussion here which is recorded and open. This is why I invited others to review this issue.
Again I state my case:
Notifications volume control is used for volume of notification pushed to the phone.
Music, video, games & other media (this is how it shows on my phone) is used for the most of the rest sounds (except ringer) and the user would expect to use this to control the sound of the non-notification sound of Habitica.

vIiRuS:
After some internal discussion we decided to not merge this in right now. We feel like the way the sounds are used don't necessarily go against the documentation, since they are used to notify you of the completion of a task. And as said, the vast majority of our users seem to agree.

Additionally, the wording you used to "invite others to review this issue" sounded more like you wanted help convincing me, rather than looking for actual feedback.

PR #743, opened 2017-04-11
Add repeatables

TheHollidayInn:
my Habitica User-ID:

This adds:

- isDue field to habits
- Adds remote config reader
- Adds UI and value setting for Repeatables when enabled


Habitica/src/main/java/com/habitrpg/android/habitica/helpers/RemoteConfigManager.java
```diff
@@ -0,0 +1,122 @@
...
+import java.net.URL;
+import java.net.URLConnection;
+
+import bolts.Bolts;
+
+/**
+ * Created by keith holliday on 4/7/2017.
+ */
+
+public class RemoteConfigManager {
```
vIiRuS: I think this should also be cached in SharedPreferences or something like that, so that it's also available when the request might fail.

Habitica/src/main/java/com/habitrpg/android/habitica/models/tasks/Task.java
```diff
@@ -608,7 +616,10 @@ public Boolean isDue(int offset) {
     }
 
     public Boolean isDisplayedActive(int offset) {
-        return this.isDue(offset) && !this.completed;
+        if (this.isDue != null) {
+            return this.isDue;
```
vIiRuS: Shouldn't this also check for `completed`?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/TaskFormActivity.java
```diff
@@ -420,7 +480,215 @@ public void onKeyboardClose() {
...
+                Integer date = calendar.get(Calendar.DATE);
+                weekdays = " on the " + date.toString();
+            } else {
+                Integer week = calendar.get(Calendar.WEEK_OF_MONTH);
+                String dayLongName = calendar.getDisplayName(Calendar.DAY_OF_WEEK, Calendar.LONG, Locale.getDefault());
+                weekdays = " on the " + week.toString() + " week on " + dayLongName;
+            }
+        }
+
+        String summary = "Repeats " + frequency + " every " + everyX + " " + frequencyQualifier + weekdays;
```
vIiRuS: Can we make this at least somewhat translatable? We probably can't cover everything in all languages, but having it not translated at all feeld wrong.

TheHollidayInn: I'm not super familiar with translations on Android. Is there a way to inject parameters into strings?

vIiRuS: Yes, they can. `This %1$s a %2$s string`  per example for two strings. https://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling

TheHollidayInn: Awesome! Thanks

Issue #718, opened 2017-02-19
copy to clipboard in inbox does not work


Alys:
In the Inbox, when you tap on a message's menu icon, the "Copy to clipboard" option does not work. Tapping on it closes the menu but gives no other result and the message does not get copied to the clipboard.

PR #746, opened 2017-04-25
Copy clipboard at inbox list fragment. From chat list fragment

josephnw:
I hope this may solve issue #718 (Copy to clipboard from inbox)
The function is from the ChatListFragment.java 41ef950068da80e623ea7a715eb7795edbabc80d
I don't know how to integrate to same file  (since the logic is the same)
my Habitica User-ID: 147dd332-1d66-494f-b1bc-08dd90f62de2
Thank you


josephnw:
I'm sorry, I made a mistake for this commit. The correct one is on the #747

PR #753, opened 2017-05-06
Update

morphors:
pequena correção
palavra errada

my Habitica User-ID:


vIiRuS:
Hi,

This is not the place to submit localization changes. All translations are handled through transife. You can read more about that here: http://habitica.wikia.com/wiki/Guidance_for_Linguists

PR #759, opened 2017-05-16
Improvement ui task speed

ioanb7:
my Habitica User-ID: eeaa2757-fb24-49f9-afbf-1104e94bf70a

Hi,

I've made a few changes to the way snackbar works. Before, when you were checking 5 items in a row you couldn't see all the rewards you were getting because the the snackber views would not be properly added in order. And even if they were, hiding and showing and hiding and showing them one by one would've been way to slow. The way it works now is that if there is a snackbar showing on the screen now, the contents are replaced (including the color of the background).

Checkout the video at: https://youtu.be/YBk2KEsV2_A to see the difference.

The ui is still very slow.. somehow the http requests are blocking the main thread if I am correct. Virus maybe you'd have a better idea on this. Let me know as this is what I want to be tackling next.

Regards,
ioanb7.


vIiRuS:
I'm not entirely sure, but I think storing a reference to the last snackbar can cause a memory leak, since the snackbar probably has a reference to the context, that would then also possibly leak. removing the reference to the last snackbar when it is dismissed should fix this.

Other than that, looks good.

as for the UI lag, right now is not a good time to tackle that, because I am finalizing a PR that ports the database to realm which fixes most of the lag issues the app has.

ioanb7:
Good to hear. I will be fixing the memory leak soon.

PR #760, opened 2017-05-17
Switch local database to realm

vIiRuS:
So this is a big one.

The main goal of this PR was to change the database to realm and move all database interaction inside of repositories. There are some other changes that were either necessary or made sense.

biggest issue in terms of reviewability of this PR is that at one point I had a big merge conflict that required me to reformat the code, which is why the diff is larger than the actual changes. I suspect this could be fixed by doing the same reformatting operation on the develop branch, but I'm not entirely certain.

The switch to realm brings huge performance improvements and now most screens show instantly (including data) instead of taking 1+ seconds to load. 🎉 
Moving all database interaction into repositories allows for simplified code in the views, which is a first step in removing the apps logic from the views.

The PR is also not completely finished and still has some issues
- [ ] moving tasks can only be done one by one when moving the task up. moving it down works fine.
- [x] tasks, checklist items and tags deleted on the website might not be removed 100% correctly in the app.

If other users that test this PR/build find any other issues, please add them as a comment using the markdown notation to create a list of checkboxes and have one list item per issue. Also please check if the issue you found has already been added, to make things easier.


olson5504:
- [ ] Challenge tabs appeared in habit list after leaving challenges section and pull to refresh habits
- [ ] Skills not displaying
- [ ] Challenge filter dialogue has seen better days
- [ ] No way to pause/unpause dailies+decaying habits 
- [ ] Party leader/description/title not displaying on party page (fixed after navigating away and back)
- [ ] No pull to refresh on party page
- [ ] Quest boss not showing HP, artwork clipping

olson5504:
- [ ] When trying to log out of app, button doesnt respond and app will just crash

olson5504:
- [ ] After uninstalling the app and reinstalling, I can't launch it without crashing ;-;

olson5504:
- [ ] Party members not displaying
- [ ] Event chat messages look very strange
- [ ] Can we make the 'Please read out community guidelines before posting' message act differently? like a dismissable banner above the write a chat message bar? Its pretty annoying to see it under the chat bar forever. Tapping the banner could take them to the guidelines and dismiss it.

olson5504:
- [ ] Tried to create a new account, got to the screen thats supposed to have justin but it was just white with the bottom bar. tapped 'next' anyways, crashed. when launched again it put me in habits with default avatar and default tasks, no walkthrough.

olson5504:
- [ ] Can't create new tasks. Tap FAB, tap task type, create it, save.. puts me back into task list with the FAB still open and no new task. Even after refresh, no task shows up.

PR #780, opened 2017-06-08
Update studio

maiatoday:
Here are all the build file changes needed to run the project in Android Studio 3. This PR does not have to be merged if everybody is still using the stable version of Android Studio, however build times seem quicker for dev builds and there are new performance tools in Studio 3. I used this guide https://developer.android.com/studio/preview/features/new-android-plugin-migration.html

We no longer need neenbedankt annotation processor as there is an annotation processor as part of the tools.

We also do not need retro lambda in with the update.

my Habitica User-ID: c34cbae3-a66f-4211-884f-0c90d48d71fc


vIiRuS:
I don't think we should switch to AS3 yet. the last time I tried it, it did not seem entirely stable yet. I'd either wait till the final release or at least until maybe a RC update that is stable.

maiatoday:
no problem, at least we have the changes we need here when we decide to update. There are quite a few gradle adjustments. I am playing with both versions so I'll keep this PR up to date as the new libs come out. I'll also fix merge conflicts a it goes along.

It is useful for people who want to use the profiling tools.

PR #781, opened 2017-06-08
Issue287/no activity ref in application

maiatoday:
Assorted memory leak fixes:
The biggest one was removing the reference to an activity from the application. Multiple leak canary leaks were appearing because of this.  I reran all of the UseCase tests and the PopupNotificationManager tests and they passed after the change.

Also I made the RemoteConfig not be a singleton and rather replaced with just method calls. From the existing code a singleton isn't needed. I left some comments in there because I don't understand the history or intention of this so maybe we want it back.

I moved the eventbus register and unregister into the basefragment so we know it is always called.

This does not close Issue 287. There are still some more leaks I am seeing with Leak Canary but I it is looking much better.

I tested by putting breakpoints in the code I changed and going to that part of the app manually. This does not mean I haven't broken anything. We probably need to do a thorough test.

my Habitica User-ID: c34cbae3-a66f-4211-884f-0c90d48d71fc

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -306,6 +306,5 @@ tasks.whenTaskAdded { task ->
 
 check { findbugs { skip true } }
 
-apply plugin: 'com.android.application' //or apply plugin: 'java'
 apply plugin: 'me.tatarka.retrolambda'
 apply plugin: 'com.google.gms.google-services'
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/RemoteConfigManager.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/RemoteConfigManager.java
@@ -28,29 +28,27 @@ import bolts.Bolts;
 
 public class RemoteConfigManager {
 
-    private static RemoteConfigManager instance;
-    private Context context;
+  //  private static RemoteConfigManager instance; //This reports a leak in leak canary, it is possible that we don't need a singleton as this class loads settings on construction into preferences. It is the only use it is never used after this again as the only other public method is static
     private static Boolean enableRepeatbles = false;
-    private String REMOTE_STRING_KEY = "remote-string";
+    private static String REMOTE_STRING_KEY = "remote-string";
 
-    private RemoteConfigManager(Context context) {
-        this.context = context;
-        loadFromPreferences();
-        new DownloadFileFromURL().execute("https://s3.amazonaws.com/habitica-assets/mobileApp/endpoint/config-android.json");
+    public static void loadConfig(Context context) {
+        loadFromPreferences(context);
+        new DownloadFileFromURL().execute(context, "https://s3.amazonaws.com/habitica-assets/mobileApp/endpoint/config-android.json");
     }
 
-    public static RemoteConfigManager getInstance(Context context) {
-        if (instance == null) {
-            instance = new RemoteConfigManager(context);
-        }
-        return instance;
-    }
+//    public static RemoteConfigManager getInstance(Context context) {
+//        if (instance == null) {
+//            instance = new RemoteConfigManager(context);
+//        }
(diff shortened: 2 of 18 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/BaseFragment.java
```diff
@@ -133,5 +108,32 @@ public String getDisplayedClassName() {
         return this.getClass().getSimpleName();
     }
 
+    private class TutorialStepTransactionListener implements TransactionListener<TutorialStep> {
```
vIiRuS: this can be a static inner class, right?

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/RemoteConfigManager.java
```diff
@@ -28,29 +28,27 @@
...
-            instance = new RemoteConfigManager(context);
-        }
-        return instance;
-    }
+//    public static RemoteConfigManager getInstance(Context context) {
+//        if (instance == null) {
+//            instance = new RemoteConfigManager(context);
+//        }
+//        return instance;
+//    }
```
vIiRuS: if this isn't needed anymore it should be removed instead of commented out.

Habitica/build.gradle
```diff
@@ -306,6 +306,5 @@ tasks.whenTaskAdded { task ->
 
 check { findbugs { skip true } }
 
-apply plugin: 'com.android.application' //or apply plugin: 'java'
```
vIiRuS: why was this removed? Isn't it needed?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/BaseFragment.java
maiatoday: it's the static inner classes that keep a reference to the outer class if they remain after the activity goes away

Habitica/build.gradle
maiatoday: it was duplicated

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/BaseFragment.java
vIiRuS: I think it's the other way around? non-static inner classes have a reference to the outer class, but static ones don't.

maiatoday: sorry misread, anonymous inner classes make the leak. Let me revisit the code.

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/RemoteConfigManager.java
maiatoday: ok I'll remove

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/BaseFragment.java
maiatoday: Static innerclasses don't keep a reference to the object, they belong to the class. Making TutorialStepTransactionListener a static inner class will remove the reference however it then requires tutorialText, tutorialTexts and tutorialCanBeDeferred to be static variables as they are needed inside the static innerclass. This means these three variables are then shared among all fragments that derive from BaseFragment and they also become GC roots.The original version (non-static inner class) made the leak canary fault go away. I will change to a static inner and retest with leak canary and check again. Will update with more info when I have it.

maiatoday:
I am going to close this pull request. I will redo the changes on the latest develop branch as many things have changed. I will make a new pull request when that has been done.

PR #790, opened 2017-07-16
fix git ignore and execute file copies

TroubleMakerBen:
my Habitica User-ID:


TroubleMakerBen:
Apologies for this. I was testing something.

PR #800, opened 2017-08-14
Update Attributes strings

### LINKED ISSUE #796: task attributes have old names (physical, mental, social, other)
When you edit a task in the app, the Attributes drop-down menu contains the old names for the attributes (e.g., Physical) instead of the names we use now on the website (Strength, etc).

Physical = Strength
Mental = Intelligence
Social = Constitution
Other = Perception

_(I had thought we already had this recorded so this might be a duplicate. I can't find the old one though. EDIT: cTheDragons has confirmed this happens in the iOS app too. Logged here: https://github.com/HabitRPG/habitica-ios/issues/476)_

MrTequila:
Fixes #796

Updates Attributes strings:
'physical' - 'Strength'
'mental' - 'Intelligence'
'social' - 'Constitution'
'other' - 'Perception'

my Habitica User-ID:  #MrTequila

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -110,10 +110,10 @@
     <string name="actions">Actions</string>
 
     <string name="attributes">Attributes</string>
-    <string name="physical">Physical</string>
-    <string name="mental">Mental</string>
-    <string name="social">Social</string>
-    <string name="other">Other</string>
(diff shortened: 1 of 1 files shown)


vIiRuS:
Thanks for this!

You actually posted your profile name. The user name can be found in the API Settings on the website which you can access through the menu in the top right or the Account Details in the app.

Issue #765, opened 2017-05-28
The Shops->Quests menu shows wrong unlock methods


StefanoTrv:
In the Shops->Quests menu, all the quests that can't be purchased with gold or gems have the same string next to them that explains how they can be unlocked: "Unlock by inviting friends". This is correct for only "The Basi-List" quest, while the others have different unlock methods.

These are the quests in the website:
![quests](https://cloud.githubusercontent.com/assets/27898124/26530419/be18145a-43d4-11e7-8632-2bb1c0ddac71.PNG)

And this is the shop in the app:
![screenshot_20170528-181955](https://cloud.githubusercontent.com/assets/27898124/26530435/d2d77afc-43d4-11e7-82c5-3bd802d1a68d.png)

![screenshot_20170528-182002](https://cloud.githubusercontent.com/assets/27898124/26530437/d71c7f72-43d4-11e7-8c3b-7357e92aa66a.png)

LunarGem:
I noticed this, I could attempt to fix it if someone hasn't already started.

SabreCat:
This will require some changes on the main Habitica repo. Basically, we need to define a new unlock method, and ensure that Android can handle it. (We originally did have separate unlock methods defined in the content files, but because Android didn't have cases for them, it crashed the app.)

LunarGem:
Is there an issue for this in the main repo? I could still try to do it.

SabreCat:
Looks like we didn't, but I've added one! See the reference above.

LunarGem:
Can someone tell me where the code that `com.habitrpg.android.habitica.R` refers to is? I'd like to make sure that my returns are correct instead of just guessing and copying from the other one.

vIiRuS:
The strings are in the res/values/strings.xml files.

LunarGem:
I'm not finding the strings I'm looking for in that file. Should I add them or am I just overlooking them? @vIiRuS

vIiRuS:
Which strings exactly are you not finding?

If I understood it correctly, the server will respond with the translated unlock text, right? In that case for the app the best solution would probably be to change the seitch statement to return the text from the field if it is not found in the switch statement. So instead of returning 0 it returns the unlock condition directly. Since in the future that would already be the translated text. That way we wouldn not have to coordinate an app update with the server change.
In the meantime additional cases can also be added to the switch statement to handle that if needed.

LunarGem:
[...]

PR #803, opened 2017-09-22
Add correct unlock conditions to shop

LunarGem:
Not done on a working dev environment, so someone ought to check and see if it blows up...
Partial fix for #765 (will need better implementation, this is a bugfix)
Tada! The Rogue Lady comes through on her promises!
my Habitica User-ID:
9e9c1391-f10c-4e07-9e72-c582daabb3a5


vIiRuS:
Thanks! I noted this as a contribution in your habitica profile

PR #811, opened 2017-10-10
Add missing property "STAGING_KEY" for builds on example

caipivara:
The current setup needs a the variable `STAGING_KEY` on the `BuildConfig` class but it not documented on the example.

This PR try to:
- Fix `STAGING_KEY` doc issue.
- Fix each project use different sdk version setup ending up in conflicts with travis setup sometimes if not updated correctly, so this makes each module to use the same sdk version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,10 @@ android:
     - tools
 
     # The BuildTools version used by your project
-    - build-tools-25.0.3
+    - build-tools-26.0.1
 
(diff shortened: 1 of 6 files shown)


caipivara:
@vIiRuS not sure what value should have the `STAGING_KEY` for travis builds, I added a random one.

vIiRuS:
This value actually isn't needed anymore. I'll remove the key completely.

PR #838, opened 2017-10-31
Don't show like button in messages

### LINKED ISSUE #816: Don't show +1 button for inbox messages
Inbox messages can't be liked, so the button should be hidden.

aleien:
Fixes #816

This fix hides "like" button in messages, but keeps it in Tavern chats.

Habitica User-ID: 8eea96e0-042c-44f4-b589-e3177f59bcdd


vIiRuS:
Thanks!

Alys:
Just checking something... this does still allow the like button to appear in guilds and parties, right?

PR #883, opened 2017-11-15
fixes #882

### LINKED ISSUE #882: Misleading message when deleting an account
When deleting an account, user gets a message that reads: "Are you sure? This will delete your account forever, and it can never be restored! You will need to register a new account to use Habitica again. Banked or spent Gems will not be refunded. If you\'re absolutely certain, type your password into the text box below."
User shouldn't type their password there - they should type "DELETE". Typing the password doesn't work, and the box isn't suitable for typing in passwords - it's a normal text box, the password is visible and stored in phone's dictionary.

sarna:
fixes #882


vIiRuS:
This unfortunately doesn't fix the problem. After asking the server devs, this is the behavior:

Case 1: User has local authentication with username and password: The app should prompt for the password (in this case, the EditText should be changed to be a secure input)

Case 2: User has google or facebook as authentication: The app should prompt them to enter DELETE instead. In this case a secure input is not needed.

sarna:
Aw, I thought it'd be that simple. Do you want me to fix it tomorrow or to leave it up to you guys? I don't know the codebase, it might get fixed quicker without me, actually.

PR #920, opened 2018-01-11
don't show hourglass or gem count for other players. fixes #911.

### LINKED ISSUE #911: 0 gems appears in all player's profiles
When you view the profile for yourself or other players, it contains a gem count, but the count is always 0.

The number of gems for players other than yourself should never be shown for privacy reasons, so it's correct that the actual count isn't shown, but it's misleading that the count is displayed as zero.

There's no privacy problem with displaying your own gem count in your own profile of course, but I suggest for consistency that the gem count isn't shown there either.

kemirea:
my Habitica User-ID: c5a52ba6-2980-41a1-b444-f20009f5d13b


Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/CurrencyView.kt
```diff
@@ -29,12 +29,15 @@ class CurrencyView : android.support.v7.widget.AppCompatTextView {
                 attrs,
                 R.styleable.CurrencyViews,
                 0, 0)
-        lightBackground = attributes?.getBoolean(R.styleable.CurrencyView_hasLightBackground, true) ?: true
+        lightBackground = attributes?.getBoolean(R.styleable.CurrencyView_hasLightBackground,
+                true) ?: true
+        visibility = GONE;
```
kemirea: I set visibility to GONE when the view is first created so that the UI does not show hourglass, gem, and coin icons before these values are loaded. I'm not very familiar with Kotlin, so if anyone knows of a way to do this without having to add it to both constructors, I'd really appreciate it.

vIiRuS:
Thanks!

I noted this as a contribution in your Habitica profile

PR #928, opened 2018-01-29
Optimize image resources with ImageOptim (with lossless algorithms) #1

caipivara:
Hi

I was just curious to see how much can [ImageOptim](imageoptim.com) can save and it results that it removed 5.3 MB from 5.9 MB (99.5%).

This may help to have a smaller apk file, small and faster memory consumption of the app at runtime.

screen shot 2018-01-29 at 11 23 45 am

https://imageoptim.com/mac


vIiRuS:
Thanks for the changes!

PR #997, opened 2018-07-04
refactor MainActivity onPause()

garritfra:
my Habitica User-ID: 6f894e69-01d1-45d1-89ee-587a596ff379

This might be very trivial, but following Curly's Law (Do One Thing), I found it would be better to refactor multiple calls to a seperate function. What do you think?


phillipthelen:
Looks good! Thanks!

I noted this as a contribution in your Habitica profile.

PR #1036, opened 2018-08-22
Skills tasks list emoji/markdown support

cvuorinen:
I abstracted the markdown parsing a little bit instead of just copy pasting the entire observable sequence like in my last PR. Is it OK to have this kind of functionality in the `Task` model and `MarkdownParser` helper, what do you think @phillipthelen ?

### DIFF
--- a/Habitica/res/layout/skill_task_item_card.xml
+++ b/Habitica/res/layout/skill_task_item_card.xml
@@ -21,13 +21,13 @@
                 android:paddingEnd="16dp"
                 android:paddingBottom="20dp"
                 android:paddingTop="20dp">
-                <TextView
+                <net.pherth.android.emoji_library.EmojiTextView
                     android:id="@+id/titleTextView"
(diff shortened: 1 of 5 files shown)


TheHollidayInn:
Looking good! 

I'd suggest either creating a TaskExtension or having a separate class for the markdown parsing. We should keep business logic off of the model.

cvuorinen:
That was my initial thought also, but since we need to assign the parsed values to the model (`parsedText` and `parsedNotes`) I thought it might be ok in the model (+ it already had logic related to markdown parsing in there).

TheHollidayInn:
I'll let Philip weigh in on the decision; I'm just giving a quick note.

But to your comment, I'm not sure we need to assign the variables to the model (looks like they are only used locally in the function/closure scope).

phillipthelen:
The models aren't great right now anyway. I'd merge this in now and then revisit getting the business logic out of all the models at a later point.

PR #1054, opened 2018-10-08
used correct serialized name for disabledClasses (#983)

### LINKED ISSUE #983: Enable Class System should not say that it will cost 3 Gems
When you are opted out of the class system, if you go to settings and click "Enable Class System", you see a message saying it will cost you 3 Gems. The expected behaviour is that you get directly into the "Select Class" screen with no message about a Gem cost.

The Gem cost message should appear only when you have previously selected a class and are not opted out of the class system.

ashleyng:
Fixes #983.  Correctly deserialize the `diabledClasses` information.

I'm not too familiar with Realm, let me know if there is anything else that needs to be done since I did add to the model's file

my Habitica User-ID: 09229c96-d005-4f97-b9ca-d8a7d84718b3

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/models/user/Preferences.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/models/user/Preferences.kt
@@ -15,6 +15,7 @@ open class Preferences : RealmObject(), AvatarPreferences {
     private var hair: Hair? = null
     var suppressModals: SuppressedModals? = null
     private var costume: Boolean = false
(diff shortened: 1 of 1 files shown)


phillipthelen:
Good find!

This change doesn't actually affect realm, since it only changes the name that gson uses for deserializing. However changes to realms schema are generally always OK, since we configured it to delete and recreate the database if realm detects a migration would be needed. This way we don't run into migration issues and since all user data is stored on the server, the only downside is that the first launch after updating takes a bit longer since the data is loaded first.

Issue #1056, opened 2018-10-13
Buffs not displayed in Stats page


cvuorinen:
Currently buffs are not showing in the Stats page. I think it used to work previously, but not sure when it broke. Going to the profile page (througj party member list) displays buffs correctly (although on that screen they are called "Boosts" which is a bit confusing). Here are screenshots:

![screenshot_20181013-151801__01](https://user-images.githubusercontent.com/1764167/46905413-007be480-cefc-11e8-8c5d-21b272839010.jpg)

![screenshot_20181013-151842__01](https://user-images.githubusercontent.com/1764167/46905416-070a5c00-cefc-11e8-908d-43d5ebfc15e9.jpg)

PR #1060, opened 2018-10-22
Fix stats page buffs

cvuorinen:
Fixes: #1056 

Don't know that much how those proxy objects work, so not sure if accessing those properties straight without calling those getters should work or not, but this fixes the problem. Also changed the "Boosts" text label in the profile view to "Buffs" for consistency.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.kt
@@ -378,7 +378,7 @@ class FullProfileActivity : BaseActivity() {
         val buffs = stats.buffs
 
         addAttributeRow(getString(R.string.profile_allocated), stats.str?.toFloat() ?: 0f, stats._int?.toFloat() ?: 0f, stats.con?.toFloat() ?: 0f, stats.per?.toFloat() ?: 0f, true, false)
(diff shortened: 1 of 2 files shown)


phillipthelen:
Have you also checked with the most recent version on the develop branch if it doesn't work correctly there? I was not able to reproduce the issue and my buffs display correctly.

cvuorinen:
I updated my develop branch just few days ago, but I will check again.

cvuorinen:
Hmm, there were a few commits I didn't have in upstream/develop but now that I merged those in I get a bunch of Gradle dependency errors (like `Failed to resolve: play-services-ads` etc.) when I did  "Sync project with Gradle files" in Android studio. Anything else I need to do?

phillipthelen:
Are you using Android Studio 3.2? The project was updates for that, so if you haven’t updated yet that might be the issue.

--  
Phillip Thelen
Sent with Airmail

cvuorinen:
I have 3.2.1 which is the latest. I had some plugin updates when I ran check for updates, but updating those did not fix the issue. Now it says "you have the latest version of Android studio and plugins" but the issue still remains the same.

cvuorinen:
Ah, changing `google()` above `jcenter()` in `build.gradle` fixed the issue (as instructed here https://stackoverflow.com/a/50774794/225075).

cvuorinen:
So I got it working with latest from develop, but the issue still remains. Buffs not shown on develop branch and these fixes make it work. Don't really know how it works differently for you, but I guess it couldn't hurt accessing the buff values through those getter methods, since that's how they are used in the FullProfileActivity class?

Issue #1085, opened 2018-12-02
Some of the text on the bottom menu is cut out and not visible.


CSEKimDoYeon:
When language is set to Korean, some of the text corresponding to each menu is not shown in the bottom bar tab of the main menu.
![default](https://user-images.githubusercontent.com/33437652/49341334-54e74900-f68f-11e8-9839-1c0709369d19.PNG)

PR #1086, opened 2018-12-02
#1085 Some of the text on the bottom menu is cut out and not visible.

CSEKimDoYeon:
When language is set to Korean, some of the text corresponding to each menu is not shown in the bottom bar tab of the main menu('일일과제', '해야할일').
So, I adjusted the value of "bb_height".

my Habitica User-ID:hb-4p3td37tlbp5iiwt8


CSEKimDoYeon:
The images before modifying 'BottomBar' are as follows. As you can see, some of the text at the bottom of the tab menu looks clipped.

default

Therefore, I think we can solve the problem by adjusting the height of the 'BottomBar' a little bit as follows.

default


If there's a better solution, I'd like to hear your feedback. Thank you.

Issue #1084, opened 2018-12-02
There are missing entries in language settings.


CSEKimDoYeon:
Settings-APP SETTINGS tab does not contain 'Korean' Settings tab. If the user wants to change to a different language and then set it back to Korean, there is no way to set it.
![kakaotalk_20181202_221640214](https://user-images.githubusercontent.com/33437652/49340205-01b9ca00-f680-11e8-87a8-e07fddec607e.jpg)

PR #1091, opened 2018-12-07
#1084 There are missing entries in language settings.

CSEKimDoYeon:
Settings-APP SETTINGS tab does not contain 'Korean' Settings tab. If the user wants to change to a different language and then set it back to Korean, there is no way to set it.

I read your comment well. Apart from modifications to the typo, it is not possible to change the language to Korean because the menu tab does not have the 'Korean' tab when using the existing Release version. Therefore, it would be better to add the 'Korean' tab to the menu tab.

my Habitica User-ID:hb-4p3td37tlbp5iiwt8

### DIFF
--- a/Habitica/res/values/values.xml
+++ b/Habitica/res/values/values.xml
@@ -59,6 +59,7 @@
 
     <string-array name="Language">
         <item>English</item>
+        <item>한국어</item>
         <item>Български</item>
         <item>Deutsch</item>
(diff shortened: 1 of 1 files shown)


CSEKimDoYeon:
Since the details of the previous 'Pull Request' were included in the details of the 'commit' that were not needed, I will send only those parts.

PR #1097, opened 2019-01-03
Fixes toolbars showing class names (#1095)

### LINKED ISSUE #1095: Toolbar showing fragment class names
I was just looking into issue #1078, and noticed that certain fragments show the name of their class in the toolbar. 

A few places this can be reproduced:
- Opening an existing chat with another user will show "InboxMessageListFragment" in the toolbar.

- Opening the stable and clicking on a pet will show "PetDetailRecyclerFragment" in the toolbar.

- Opening the stable, navigating to the mounts and clicking on a mount will show "MountDetailRecyclerFragment" in the toolbar.

I've identified the cause as [navigation.xml](https://github.com/HabitRPG/habitica-android/blob/develop/Habitica/res/navigation/navigation.xml#L186-L218) - it looks like some of the `android:label` attributes contain the names of the fragments rather than string resources with the proper names. 
The Pet and Mount fragments should just require a string resource, but the InboxMessageListFragment may take a bit more time to solve (again referencing #1078) as the text in the toolbar will need to be dynamically set to the name of the other user.

This file is part of @phillipthelen's recent navigation improvements, so I decided to create an issue for it to avoid a potential conflict.

debashisbiswas:
This issue fixes #1095 by using string resources, rather than class names, for certain `android:label` properties in `navigation.xml`.

my Habitica User-ID: 23ebecac-146a-4d6e-ab1d-3a94da8fd807

### DIFF
--- a/Habitica/res/navigation/navigation.xml
+++ b/Habitica/res/navigation/navigation.xml
@@ -115,7 +115,7 @@
     <activity
         android:id="@+id/gemPurchaseActivity"
         android:name="com.habitrpg.android.habitica.ui.activities.GemPurchaseActivity"
-        android:label="GemPurchaseActivity" />
+        android:label="@string/gem.purchase.toolbartitle" />
     <fragment
         android:id="@+id/newsFragment"
         android:name="com.habitrpg.android.habitica.ui.fragments.NewsFragment"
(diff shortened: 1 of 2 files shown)


phillipthelen:
Thanks for the changes! Looks good to me.

I'll note this as a contribution in your habitica profile.

PR #1102, opened 2019-02-03
Fix text of delete account confirmation

### LINKED ISSUE #882: Misleading message when deleting an account

ajsviado:
Should ask user to input password on a secure input text box when
user has a local authentication. Otherwise ask the user to input
the text DELETE on a plain text box.

Fixes #882

my Habitica User-ID:ajsviado


phillipthelen:
Thanks for the changes!

Issue #794, opened 2017-07-24
Global Notification Support


LunarGem:
Not sure if I'm using the right language for this, but I wanted to work on having the same notification support on the app as on the website. The inbox notifications do not currently appear to be working, and having guild and party notifications would be nice as well. I would do the code for this if it were something we were looking to implement.

saraolson:
We'll be adding something like this after web finishes their notification rehaul and we finish a menu rehaul

cvuorinen:
Has there been any progress/discussion/design on this issue? It's the main reason I keep switching between the mobile app & website, so would be nice to get the notifications in the mobile app.

cc @phillipthelen

saraolson:
@cvuorinen there hasn't been any progress on this feature just yet, though I believe the notification rehaul was completed and we did the part of the menu update we had planned. 

Are you still interested? If so, I'll discuss the me shifting my priorities to making designs with the team so you'd have something to work off of.

cvuorinen:
Yep, I'm still interested. Especially now since the website updated the menu and notifications is not usable on mobile through the website anymore (though I do hope it gets fixed soon).

saraolson:
@cvuorinen great! alright, i'll work on adapting designs to mobile and give you links to the mockup by the end of the week hopefully with some notes from us on how it we're thinking it might behave. after that, we'll probably just ask for you to brief us on your plan of attack then let you go at it!

saraolson:
Alright! Designs are finished and approved. We use Zeplin to handle our mockups. Can you give me an email address you'd like to use to access the Zeplin project board so I can invite you?

After gaining access, mockups can be found here: [Zeplin Designs](https://zpl.io/anpdePQ)

Our notes on which notifications to include in the mobile implementation and what tap areas should do can be found on this google sheet for reference: [Mobile Notification Breakdown](https://docs.google.com/spreadsheets/d/1I76AKl9chH_FHTpcjpqDNcsCat4mNdPp0gtPSGYN130/edit#gid=0)

Let us know if you have any questions now or throughout the process :)

saraolson:
oops i forgot to tag you in my previous comment, @cvuorinen. designs are ready!

cvuorinen:
@saraolson carl.vuorinen@gmail.com

saraolson:
@cvuorinen thankya, sent you an invitation

cvuorinen:
[...]

PR #1116, opened 2019-03-17
Global Notification support

cvuorinen:
Work In Progress on issue  #794 (Global Notification support)

Wanted to open this pull request to share the work I have done and to ask some questions + feedback/discussion.

Current status: notification icon added to the drawer with notification count badge, click opens notification view (which also supports pull to refresh), displays a list of notifications if there are any, displays the "You’re all caught up" text otherwise. Currently supported notifications: new chat message, Bailey update, unallocated stat points.

Todo:
- [x] Dismissing notifications
- [x] Marking notifications as seen
- [x] Click handlers for notifications
- [x] Rest of the "dismissable" notifications
- [ ] Notifications with Accept/Decline buttons
- [ ] Translations for notification texts


Habitica/src/main/java/com/habitrpg/android/habitica/models/notifications/GlobalNotification.java
```diff
@@ -0,0 +1,21 @@
...
+import io.realm.annotations.PrimaryKey;
+
+/**
+ * Represents Habitica "Global notifications", i.e. the notifications about chat messages
+ * (in Guilds and Party), Party & Quest invitations, unallocated stat points etc.
+ *
+ * These are different from other kind of notifications, such as Push notifications and
+ * Popup notifications.
+ */
+public class GlobalNotification extends RealmObject {
```
cvuorinen: Not sure what would be good name for the model since there already exists a class called [Notification](https://github.com/HabitRPG/habitica-android/blob/1.6.1/Habitica/src/main/java/com/habitrpg/android/habitica/models/Notification.java#L12). Named it `GlobalNotification` for now but open for better suggestions.

Habitica/res/layout/activity_notifications.xml
```diff
@@ -0,0 +1,54 @@
...
+                    android:id="@+id/notification_items"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="vertical"
+                    android:divider="?android:listDivider"
+                    android:showDividers="middle">
+                </com.habitrpg.android.habitica.ui.views.PaddedLinearLayout>
+            </androidx.core.widget.NestedScrollView>
+        </androidx.swiperefreshlayout.widget.SwipeRefreshLayout>
+    </FrameLayout>
```
cvuorinen: I'm not that familiar with all this layout stuff so any feedback would be welcome. I made it an activity, but don't know if just fragment would have been better.

Habitica/src/main/java/com/habitrpg/android/habitica/models/notifications/GlobalNotification.java
phillipthelen: GlobalNotification sounds good to me for now. Do you have a plan yet for how to get the different notification types into realm? Since the data will come from multiple sources (Some of them are already stored through the user object) and with storing it in realm, you then also have to take care of removing old data. Could this maybe also be implemented as more of a viewmodel? Subscribing to p.e. the user object and then dynamically creating these notification objects from the data stored there using RxJava should work pretty well. (Please take this as a suggestion and not a demand. if you already have a plan that you think works good/better, that's good too!)

Habitica/res/layout/activity_notifications.xml
phillipthelen: An activity is fine by me.

Habitica/src/main/java/com/habitrpg/android/habitica/models/notifications/GlobalNotification.java
cvuorinen: Thanks for the suggestion, I'll take a look into it. Don't have a concrete plan yet but I've been looking into realm and was also thinking it might not be best way. Just wanted to get something working for now, so that I can get something visible to show up on screen :smiley:

cvuorinen:
Here is a screenshot of the current status. Added a checklist to the PR description to indicate what work is still to do. Any feedback is welcome. I also posted few questions yesterday to Zeplin.

Screenshot_1554839006

Habitica/res/values/strings.xml
```diff
@@ -860,5 +861,12 @@
     <string name="discover">Discover</string>
     <string name="damage_paused">Damage paused</string>
     <string name="preference_push_important_announcements">Important Announcements</string>
+    <string name="no_notifications_title">You’re all caught up!</string>
+    <string name="no_notifications_text">The notification fairies give you a raucous round of applause! Well done!</string>
+    <string name="dismiss_all">Dismiss All</string>
+    <string name="new_bailey_update">New Bailey Update!</string>
+    <string name="new_msg_guild"><![CDATA[<b>%1$s</b> has new posts]]></string>
+    <string name="new_msg_party"><![CDATA[Your Party, <b>%1$s</b>, has new posts]]></string>
+    <string name="unallocated_stats_points"><![CDATA[You have <b>%1$s unallocated Stat Points</b>]]></string>
```
cvuorinen: It's ok to use `Html.fromHtml` with these, or is there some better way to achieve same thing?

Habitica/src/main/java/com/habitrpg/android/habitica/models/notifications/GlobalNotification.kt
```diff
@@ -0,0 +1,64 @@
...
+    var newChatMessageData: NewChatMessageData? = null
+    var unallocatedPointsData: UnallocatedPointsData? = null
+
+    // Workaround for Realms lack of polymorphism
+    fun getData(): GlobalNotificationData? {
+        return when (type) {
+            NotificationType.NEW_STUFF.type -> newStuffData
+            NotificationType.NEW_CHAT_MESSAGE.type -> newChatMessageData
+            NotificationType.UNALLOCATED_STATS_POINTS.type -> unallocatedPointsData
+            else -> null
```
cvuorinen: Not sure if there is better way to handle the `data` property of the notifications since it's different type for each notification (as Realm doesn't support polymorphism AFAIK). Suggestions/feedback is welcome. If I'm doing something stupid here, feel free to tell me :smiley:

Habitica/res/values/strings.xml
phillipthelen: I haven't really found a better way to do this yet, so jeah, that's ok.

Habitica/src/main/java/com/habitrpg/android/habitica/models/notifications/GlobalNotification.kt
phillipthelen: I'd maybe change the access level of the actual fields to something more restricted (since you're probably not supposed to interact with them directly), but while it does look strange it's probably the best solution given realms restrictions here!

phillipthelen: Looking at this again, I noticed that none of the Data objects have a primary key with the ID of the notification. This will be needed (or at least very helpful) for deleting old data from the database.

cvuorinen: Yeah, I'm actually currently looking into avoiding saving these to Realm at all. I realized at some point that the already existing `Notification` model is actually kind of same thing, just that it models one specific type of notification instead of all the different ones, and the `apiCallTransformer` in `ApiClientImpl` is something that handles these from each request. So I'm looking at if I could model all the notifications using similar approach as I did with `GlobalNotification` and then might hook into the apiCallTransformer as well with this new stuff.

cvuorinen:
So I made quite a lot of changes in the way the model is implemented in the previous commits. Now not saving anything to realm and using the apiCallTransformer to make it work with notifications from each request.

Now trying to implement the click handlers, but I need some help. Trying to call something like `MainNavigationController.navigate(R.id.partyFragment)` from the `NotificationsActivity` class does not do anything. Tried to read up on the navigation controller but didn't find anything useful yet. Is this the correct way to navigate?  Any pointer what additional steps are required to make it work?

phillipthelen:
This is probably because the navigation controllers only work for a single activity. so the MainActivityController only works for changing fragments within the MainActivity and opening activities from the MainActivity. I'm not entirely sure how to handle this case best, but one idea would be to finish the NotificationsActivity with a result, that would then be handled by the MainActivity (opening the fragment that is passed through the ResultIntent). This would mean that if the user taps the back button, they would not get back to the notification view however. I'm not what UX android users would expect here.

cvuorinen:
Thanks, I got it working using a result intent.

cvuorinen:
Btw, I've also been looking into running the unit tests but without success.

Tried just right-clicking on one test file and select Run 'SomeTest' and also from command-line with `./gradlew testProdDebugUnitTest -PdisablePreDex --stacktrace` but both give similar errors. Just asking so that I know if they are supposed to work like that (so I know the issue is with my env) or I should do something else to run them.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/viewmodels/NotificationsViewModel.kt
```diff
@@ -0,0 +1,144 @@
...
+            Notification.Type.NEW_CHAT_MESSAGE.type -> clickNewChatMessage(notification, navController)
+            Notification.Type.NEW_MYSTERY_ITEMS.type -> navController.navigate(R.id.itemsFragment)
+            Notification.Type.UNALLOCATED_STATS_POINTS.type -> navController.navigate(R.id.statsFragment)
+        }
+    }
+
+    private fun clickNewChatMessage(notification: Notification, navController: MainNavigationController) {
+        val data = notification.data as? NewChatMessageData
+        if (isPartyMessage(data)) {
+            navController.navigate(R.id.partyFragment)
```
cvuorinen: I guess there's currently no way to navigate straight to the chat view of party/guild fragment? And same with items fragment special view?

cvuorinen: Maybe could add an argument to open a specific view in the fragment? Or is there some other way to achieve it?

cvuorinen:
I think this is ready for review and it could also be merged. I'm still working on the actionable notifications but this works as is and I could make a separate PR for the rest of the changes.

saraolson:
@cvuorinen our mobile dev is out this week but i'll make sure to point this out to him for review when he's back!

cvuorinen:
@phillipthelen Hmm, we seem to have conflicted that `PopupNotificationsManager` (which I renamed to `NotificationsManager` since it handles also other notifications in this branch). Have you had the chance to check out this PR yet? Wondering what would be the best way to handle this, I guess the version you converted from java to Kotlin is mostly same functionality as before?

phillipthelen:
@cvuorinen I have not had the chance to look at the updates yet! I only got back from a short vacation on friday and only did some smaller things to catch up, but I'll go through it today. Since I caused the conflict, I'll try to resolve it myself (I think github should let me do that)

phillipthelen:
Thank you a lot for your work! I couldn't find any glaring issues, so I merged it into develop, so that it can be pushed to the beta release channel later today! (If you're not yet part of it and also want to test, there is a button to let you join it when you open the app page in the play store app).

I noted this as a contribution in your habitica profile!

cvuorinen:
Great, I'm already a beta tester so waiting for the update :+1:

I'll open a new PR soon for the remaining things. I'm already pretty much done with the actionable notifications (just added one question in Zeplin today about the quest invitation notification).

PR #1130, opened 2019-04-25
Long usernames no longer cover up likes in chat

### LINKED ISSUE #1076: Chat: Likes not visible if display name is too long
On lower density screens, display names can be so long that they push the like count off the message card. this may be especially bad with staff and mod tags

bacowan:
Fixes #1076

Long usernames in chat will now word-wrap if they are too long instead of pushing the likes off the edge of the screen.

Switched the `LinearLayout` that contains the username, moderator tag, and likes to a `TableLayout` for better control of the widths of those elements (most of what's marked as changed in `tavern_chat_item.xml` is indentation from the required `TableRow`).

Gave the `TextView` inside of `UsernameLabel` a different weight from the `ImageView` so that the tier icon isn't pushed off as well.

my Habitica User-ID: 55a3cc27-2b92-4901-8313-e194c070bc6e

Screenshot_1556588710


bacowan:
By the way, do you squash PR merges? If not I can make another PR with a cleaner commit history.

saraolson:
@bacowan hey there! brought this back to the team to discuss which we prefer and we chose the option with the display name wrapping to the next line + the shortened (mod) tag. we've gotten complaints in the past of people being unable to see their full name and since we have a charater limit on display names now, it shouldn't get too out of control.

as for your questions about squashing PRs: it depends on the PR, but we can definitely do that in this case. so no need to clean the history up yourself

bacowan:
Hey @saraolson , sounds good! I've updated the PR accordingly. Note that the mod tag will only be shortened for English right now of course. If you think it would be better for the tag to shrink dynamically with the length of the name I could look into that too, though I'm not sure off the top of my head how I'd go about it.

phillipthelen:
Thanks!

PR #1147, opened 2019-05-08
Including the option to share a challenge

### LINKED ISSUE #1118: Ability to share a Challenge
Have a 'Share Challenge' option in the overflow menu accessible through the Challenge details screen that would give users a copied link they can then paste into whatever chat service they'd like to allow friends to view and potentially join a Challenge.

This is a quick way to add the ability to share a Challenge without having to go through the implementation of inviting and notifying. It could also be useful for people who would like to share the Challenge with a larger group without having to individually notify accounts (ie social media)

bacowan:
Fixes #1118 

The overflow menu in the Challenge Details screen now includes a Share option, which opens a share sheet with the Challenge URL.

my Habitica User-ID: 55a3cc27-2b92-4901-8313-e194c070bc6e

image

### DIFF
--- a/Habitica/build.gradle
+++ b/Habitica/build.gradle
@@ -63,6 +63,7 @@ dependencies {
     kapt 'com.google.dagger:dagger-compiler:2.22.1'
     compileOnly 'javax.annotation:javax.annotation-api:1.3.2'
     //App Compatibility and Material Design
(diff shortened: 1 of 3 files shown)


phillipthelen:
Thanks!

Issue #1153, opened 2019-05-15
Tasks: Keyboard goes away immediately when trying to adjust a streak


saraolson:
This makes it so you cant actually type. KB Will pop up for a fraction of a second then go back down on its own.

rsamborski:
Fix proposed #1158

arajago6:
Also, the adjusted streak is not getting saved for habits. For Dailies, it is getting saved though.

rsamborski:
@arajago6 Thanks for this information. When looking closer into code and API (https://habitica.com/apidoc/#api-Task-UpdateTask) it seems there is no such value as positive/negative streak for Habits. There are only a boolean fields up/down which indicate if habit is positive or negative.
 
@saraolson @phillipthelen do you know how this positive and negative streak inputs should work?

rsamborski:
@saraolson @phillipthelen @arajago6 It seems this was undocumented API. I checked the Web App to see how it works there and my fix is available in the same PR #1158

Additionally I noticed that in Web App only proper field (positive, negative or both) shows up depending on the selected Habit type. Do we have a task to make this behavior in Android similar? I can work on this in some spare time.

saraolson:
@rsamborski thanks for digging into this! as for showing the corresponding streak field based on what the habit is set to (positive, negative, or both), feel free to add that to this PR rather than creating a new issue.

rsamborski:
@saraolson sure, I am just not sure if I will be able to do any coding in upcoming 1-2 weeks. Therefore it would make sense to merge current PR and work on showing/hiding corresponding streak field separately. Nevertheless if time is not so important I can push this change altogether.

saraolson:
@rsamborski oh alright! our mobile developer is out this week so nothing will get merged quite yet but i'll direct his attention this way when he gets back to see if he wants to add that himself to roll it out all at once or split it out.

saraolson:
created a new issue #1168 to seperate out the display of the correct fields

PR #1158, opened 2019-05-19
Fixing bug #1153 Tasks: Keyboard goes away immediately when trying to adjust a streak

rsamborski:
Turned off keyboard hiding after scroll for Positive and Negative streak edittext fields.

my Habitica User-ID: tostu


phillipthelen:
Thanks!

PR #1171, opened 2019-06-05
Global Notification support - pt2

### LINKED ISSUE #794: Global Notification Support

cvuorinen:
Continuation of #1116 

Fixes #794 


cvuorinen:
I'm still waiting for an answer on one question in Zeplin (about boss quest display, should it be consistent with other places and display boss health, or display "Defeat: boss name" as in the design, but then there are no translations for the word "defeat" as it is not used anywhere else). I'm also still working on adding translations for all the new strings (that exist in website translations).

This is what it looks like ATM:

Screenshot_1559712883

cvuorinen:
I've added all the translations found in the website translations now. And if the boss name in quest notification is ok, then this is ready for review.

Habitica/res/values-de/strings.xml
```diff
@@ -790,15 +790,27 @@ Die Quest-Schriftrolle wird an den Quest-Besitzer zurückgegeben.</string>
...
+  <string name="username_level">\@%s ・Lvl %d</string>
+  <string name="verification_pet">Eines dieser Veteranen-Haustiere wartet auf Dich wenn Du die Bestätigung abgeschlossen hast!</string>
+  <string name="gift_subscription">Abonnement verschenken</string>
+  <string name="create_party">Gruppe erstellen</string>
+  <string name="notifications">Mitteilungen</string>
+  <string name="no_notifications_title">Du bist auf dem Laufenden!</string>
+  <string name="no_notifications_text">Die Mitteilungsfeen geben dir eine heftige Runde Applaus! Gut gemacht!</string>
+  <string name="dismiss_all">Alle entfernen</string>
+  <string name="new_bailey_update">Neues Update von Bailey!</string>
+  <string name="new_msg_guild"><![CDATA[<b>%1$s</b> hat neue Beiträge]]></string>
```
phillipthelen: Have you tested this? I'm not sure if the `<![CDATA[` part works?

phillipthelen: Nevermind. Just noticed that this is also in the english versions and also in the ones already merged in and it works.

PR #1225, opened 2019-09-19
Multiplatform model migration

TNychka:
This is the initial migration from the Native Android Application to Kotlin Multiplatform. There is a significant amount of code that needs to be refactored. Starting the migration with this first PR by extracting a single common business logic file `ScoreTaskLocallyInteractor.kt`, the majority of the Android models also will need to migrate to Kotlin multiplatform to make the project compile. 

Since this change is so large, the draft PR is provided to track progress and design.


TNychka:
Status:
- App does not compile.
- IOS library is mostly placeholder/TODO/not structured at all
- Android library is compatible with the common code with the exception of User.kt, Task.kt, Stats.kt. Basically anything important
- Exactly how the multiplatform library interacts with the native code hasn't been checked at all

Todo:
- Fix User, Task, Stats common code. This is likely going to require extracting most of the remaining native models.
- Compile
- Fill better placeholders into IOS code
- Check interactions with native code

Changes:
- Extracted ScoreTaskLocallyInteractor.kt to common code.
- Extracted a number of models to the common library to resolve compile time errors

TNychka:
Changes:
- Required classes have all migrated to common code. 
 TODO:
- Solve issues with resources
- Solve issues with markdown parsing
- Compile
- Clean up IOS placeholders

TNychka:
Status:
   - App compiles!
   - IOS library is mostly placeholder/TODO/not structured at all
   - Android library is compatible with the common code

Todo:
    Fill better placeholders into IOS code
    Test very thoroughly

Changes:
   Extracted the minimum set of model files required into the common library and got the scoreLocallyInteractor to function
   Changed how stats and tasks manage their resource properties. They now go through a processor in the ui.helpers which will provide the resources they require. This is due to how the shared library handles resource access.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/AvatarWithBarsViewModel.kt
```diff
@@ -65,10 +66,10 @@ class AvatarWithBarsViewModel(private val context: Context, view: View, userRepo
         avatarView.setAvatar(user)
 
         if (stats.habitClass != null) {
-            userClass = stats.getTranslatedClassName(context)
+            userClass = getTranslatedClassName(stats, context)
```
TNychka: Due to the way that the shared library handles resources (even for the native side of the shared code) the translated class name had to be extracted to a helperclass

shared/build.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-multiplatform'
 apply plugin: 'kotlin-kapt'
+apply plugin: 'realm-android'
```
TNychka: I don't understand gradle enough to know wether this file is optimal or not. I'm sure it could be cleaned up and I'll see what I can do, but it does otherwise work.

shared/src/androidMain/kotlin/com/habitrpg/shared/habitica/MultiplatformRealmLibrary.kt
```diff
@@ -0,0 +1,6 @@
+package com.habitrpg.shared.habitica
+
+import io.realm.annotations.RealmModule
+
+@RealmModule(library = true, allClasses = true)
+actual open class MultiplatformRealmLibrary
```
TNychka: Required to introduce the shared library as a source of realm models, compilation will fail otherwise. See the initialization process for the usage.

TNychka: Habitica/src/main/java/com/habitrpg/android/habitica/HabiticaBaseApplication.kt:112

Habitica/src/main/java/com/habitrpg/android/habitica/ui/AvatarWithBarsViewModel.kt
TNychka: The stats class can take context as a parameter, however access to the resources of the habitRPG project will fail. I think it might be possible to migrate the HabitRPG resources into the shared project, but I would do that as a second PR. 
In fact, I could write that branch and attempt to merge it with this one so the changes are distinct and we could review that separately

shared/src/androidMain/kotlin/com/habitrpg/shared/habitica/MultiplatformRealmLibrary.kt
TNychka: I haven't tested it out yet, but it appears that iOS is able to handle this. If I'm correct from a brief glance https://realm.io/docs/swift/latest/#configuring-a-realm this seems to indicate all you need to do is load both the models. It looks like the library configuration is required for android only

Habitica/src/main/java/com/habitrpg/android/habitica/ui/helpers/TaskTextParser.kt
```diff
@@ -0,0 +1,51 @@
+package com.habitrpg.android.habitica.ui.helpers
+
+import com.habitrpg.shared.habitica.models.tasks.Task
+import io.reactivex.functions.Consumer
+
+object TaskTextParser {
```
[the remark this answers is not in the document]
TNychka: I just learned about kotlin extensions. I'll see what I can do with those!
Originally I did this because the object itself couldn't contain any parsing logic due to dependencies. Extensions seem to be the correct solution to that problem rather than what I've done

shared/src/androidMain/kotlin/com/habitrpg/shared/habitica/models/inventory/Customization.kt
```diff
@@ -0,0 +1,90 @@
...
+                return false
+            }
+
+            return !(this.availableUntil != null && !this.availableUntil!!.after(today))
+        }
+
+    actual val isUsable: Boolean
+        get() = this.price == 0 || this.purchased
+
+    actual val path: String
```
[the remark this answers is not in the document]
TNychka: The issue with shared platform is that it doesn't allow partial definitions if the class is expected. So everything needs to be defined in the actual definition unfortunately. There are a few things I want to test but I'm not sure that there is a way around this

TNychka:
Closing in favour of an updated and improved MP PR

PR #1226, opened 2019-10-02
Kotlin Multiplatform Model Architecture Test

TNychka:
Prototype to show that the model extraction will function as intended, and to demonstrate the design that will be applied to the rest of the models moved to the shared library.

**Changes:**
Achievement.kt realm model was extracted to shared library

The only Achievements model was moved into the common library since it is rather contained.
- Compiles.
- Interacts with existing native implementations.
- Though not demonstrated in this PR, enables both the common library and native code to create and manipulate achievements objects.

**Testing:**
Only tested on Android Emulator (pixel 2). 
- Achievements tab opens.
- Can get new achievements (tested with max warrior equipment achievement)
- Can display received achievements
- Can save, load, render from local storage without server.


TNychka:
This is just a preview of the larger migration pull request. Closing
https://github.com/HabitRPG/habitica-android/pull/1225

Issue #1079, opened 2018-11-14
Todos: Drag and drop still not completely reliable


saraolson:
Tasks stay gray for awhile and if you try to reorganize too fast it jumps all around.

saraolson:
Closing for now until we hear its messing up again.

Alys:
We see this reported fairly often, not always for To-Dos.

Here's one that came in my email about an hour ago:
"You probably are already aware of this problem, but I wanted to report a bug with sorting dailies and tasks on mobile app (I’ve encountered this problem on both Android and iPhone). When trying to move a daily further down the list it only moves a couple of steps or not at all - it seems at first like it’ll move but then it jumps back. I’ve also encountered this with to-dos, but to a lesser extent and usually fixable simply by repeating the action (which does not often help with dailies)."

PR #1232, opened 2019-10-12
Fixes for todo lists drag and drop

aleien:
Fixes for issue #1079 

1) Drag'n'drop stopped working when you switch between 'Habits' and 'To-Do' tabs.
**Cause:** `allowReordering()` was called only when adapter is null, but when you switch between tabs it isn't
**Solution:** moved `allowReordering()` out of adapter nullability check
2) Item on dailies and todo's fragment are gray for a long time after movement finished
**Cause:** Color was set in bindViewHolder for dailies and todos, but not for habits
**Solution:** Removed color setting (behavior is similar to items on habits now)
3) Fast changes on lists cause items to jump around
**Cause:** Since we are using RealmRecyclerViewAdapter with autoUpdate=true, when we change task position in `clearView()`, it tried to update list visually. Hence there was an attempt to remove animations by using `skipAnimation`. But it just hid animations, while really data wasn't changed, so fast changes caused conflicts.
**Solution:** I saw there was an attempt to use custom listener to prevent such behavior, but it wasn't used in code. I removed hack with `skipAnimation` and implemented more correct approach - by not notifying adapter if movement was already tracked by ItemTouchHelper. It still not perfect, but it looks better than before. 

Also I have proposal is to rewrite drag'n'drop system completely, modifying local tasks repository in `ItemTouchHelper.onMove`, and uploading data to server same as now (after movement is finished; or maybe even write some timed sync system for that). That will eliminate visual inconsistencies much more reliable. 
But I am not sure that that would be correct to fix in this pr. Should I create an issue with this proposal?

my Habitica User-ID: 8eea96e0-042c-44f4-b589-e3177f59bcdd

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/tasks/RealmBaseTasksRecyclerViewAdapter.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/tasks/RealmBaseTasksRecyclerViewAdapter.kt
@@ -3,6 +3,7 @@ package com.habitrpg.android.habitica.ui.adapter.tasks
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
 import com.habitrpg.android.habitica.helpers.TaskFilterHelper
 import com.habitrpg.android.habitica.models.responses.TaskDirection
 import com.habitrpg.android.habitica.models.tasks.ChecklistItem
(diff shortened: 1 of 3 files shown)


aleien:
Travis build failed due to:
> Failed to install the following Android SDK packages as some licences have not been accepted.
     build-tools;29.0.2 Android SDK Build-Tools 29.0.2
     platforms;android-29 Android SDK Platform 29

phillipthelen:
Hi,

This looks good! I think the best approach would be to merge this in for now, to improve the situation immediately. And then making a new PR with your proposed change. 

I noted this contribution in your habitica profile.

PR #1267, opened 2020-02-28
Rename appropriate sections in Pets & Mounts screen

lelandhwu:
Functionality: renamed the "drop" section in pets and mounts so that it now says "standard"

Did this by adding a getTranslatedType method in Pet and Mount classes that override the the parent method in Animal class. 

Screenshot_20200228-141506

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -954,4 +954,7 @@
     <string name="unlock_level_short">Level %d</string>
     <string name="not_participating">You are not participating</string>
     <string name="quest_completed">Quest completed!</string>
+
+    <string name="standard"> standard </string>
+    <string name="drop"> drop </string>
 </resources>
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -955,6 +955,6 @@
     <string name="not_participating">You are not participating</string>
     <string name="quest_completed">Quest completed!</string>
 
-    <string name="standard"> standard </string>
-    <string name="drop"> drop </string>
+    <string name="standard"> Standard </string>
+    <string name="wacky"> Wacky </string>
 </resources>
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Animal.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Animal.java
@@ -15,8 +15,6 @@ public interface Animal {
 
     void setType(String type);
 
-    String getTranslatedType(Context c);
-
     String getAnimal();
 
     void setAnimal(String animal);
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Animal.java
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Animal.java
@@ -1,5 +1,4 @@
 package com.habitrpg.android.habitica.models.inventory;
-import android.content.Context;
 
 public interface Animal {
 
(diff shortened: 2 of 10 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Animal.java
```diff
@@ -15,6 +15,8 @@
 
     void setType(String type);
 
+    String getTranslatedType(Context c);
```
phillipthelen: I forgot that Animal is an interface. Can you try moving the `getTranslatedType` method to a [kotlin extension](https://kotlinlang.org/docs/reference/extensions.html) for the Animal interface? That way, the method would only have to be implemented once for both the pets and mounts classes.

Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/Mount.java
```diff
@@ -44,6 +45,18 @@ public void setType(String type) {
...
+    @Override
+    public String getTranslatedType(Context c) {
+        String currType = type;
+        String translation = c.getString(R.string.drop);
+
+        if (currType.equals(translation)) {
+            currType = c.getString(R.string.standard);
+        }
+
+        return currType;
```
phillipthelen: The way this currently works is that it will get the _translated_ name of `drop`. With that the comparison would not work in all cases, since if the user is using the app in p.e. german, the value of `translation` would not be `drop` anymore. Instead the comparison should just happen with the string `"drop"`, so that it works in all languages.

Can you also add a mapping for the following:
"premium" -> "Magic Potion"
"quest" -> "Quest"
"wacky" -> "Wacky"
"special" -> "Special"

With the values it maps to all being translated the way you currently have with the `standard`. The reason for also mapping the quest, wacky and special is that with this, they can also be translated into other languages.

Habitica/res/values/strings.xml
```diff
@@ -953,4 +953,7 @@
     <string name="unlock_previous_short">Finish Quest %d</string>
     <string name="unlock_level_short">Level %d</string>
     <string name="not_participating">You are not participating</string>
+
+    <string name="standard"> standard </string>
```
phillipthelen: This should be `Standard` with the uppercase first character and without the padding spaces.

Habitica/src/main/java/com/habitrpg/android/habitica/models/inventory/AnimalExtension.kt
```diff
@@ -0,0 +1,23 @@
...
+        currType = c?.getString(R.string.quest)
+    }
+    if (currType == "wacky") {
+        currType = c?.getString(R.string.wacky)
+    }
+    if (currType == "special") {
+        currType = c?.getString(R.string.special)
+    }
+
+    return currType
```
phillipthelen: This looks correct!

One small change: I would prefer it to use one if/else construct instead of multiple if. Reason being that with this, even if it has already found the correct translation it will still check all the following cases. With an if/else, the following if-checks are not evaluated once one is true.
Alternatively this would also be a good case for kotlins [when](https://kotlinlang.org/docs/reference/control-flow.html).

lelandhwu:
I was able to get this to work. What happened:

1. New tests that I was writing by default were requiring a build of the whole project pre-launch. I replaced this requirement with gradle-make-aware, as was being done with all other previous tests.

I also made changes to `getTranslatedType` so that any null contexts will return the type. I wrote a test case for this.

Something to be aware of is I ran into the following error every time I tried to build my unit tests:

```
error: cannot find symbol Assert.assertFalse(attributeRow.getIsSummary());
```

I was able to suppress this by going into `/Users/Leland/habitica-mobile/habitica-android/Habitica/src/test/java/com/habitrpg/android/habitica/helpers/UserStatComputerTest.java:76:` and removing and references of .getIsSummary(). We could consider fixing this and other tests in the future.

PR #1268, opened 2020-02-29
add matt banner to pet and mounts screen

lelandhwu:
Adds the banner for Matt in the Pets & Mounts screen.

`ItemDecorator` was used for all `viewType`s so I edited the `marginDecorator` so that banner fit the whole screen.

image


Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/inventory/StableRecyclerAdapter.kt
```diff
@@ -34,13 +39,14 @@ class StableRecyclerAdapter : androidx.recyclerview.widget.RecyclerView.Adapter<
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): androidx.recyclerview.widget.RecyclerView.ViewHolder =
             if (viewType == 0) {
-                val view = LayoutInflater.from(parent.context).inflate(R.layout.customization_section_header, parent, false)
-                SectionViewHolder(view)
+                val view = parent.inflate(R.layout.shop_header)
+                ShopRecyclerAdapter.ShopHeaderViewHolder(view)
```
phillipthelen: Unless I'm looking at the change wrong, I think this replaces all the section headers with the NPC banner? What should happen instead is that the section headers stay the same and there is a new view holder at the top of the recyclerview that has the NPC banner.

To achieve that, I would change the `getItemCount` method on the adapter, so that it returns `itemList.size + 1`, since we want that extra item for the header. And then for `getItemViewType` it would have to return a new type to distinguish the header viewholder. And places where you access the itemList would need to be offset by one, to account for the header being at the top. I hope that explanation is somewhat clear!

Habitica/src/main/java/com/habitrpg/android/habitica/ui/helpers/MarginDecoration.kt
```diff
@@ -10,6 +10,14 @@ class MarginDecoration(context: Context?) : RecyclerView.ItemDecoration() {
     private val margin: Int = context?.resources?.getDimensionPixelSize(R.dimen.grid_item_margin) ?: 0
 
     override fun getItemOffsets(outRect: Rect, view: View, parent: RecyclerView, state: RecyclerView.State) {
-        outRect.set(margin, margin, margin, margin)
+        for (i in 0 until parent.childCount) {
+            val view: View = parent.getChildAt(i)
+            val position: Int = parent.getChildAdapterPosition(view)
+            val viewType: Int? = parent.adapter?.getItemViewType(position)
+            if (viewType == 2) {
+                outRect.set(margin, margin, margin, margin)
```
phillipthelen: The code itself looks good and I like the solution! Only thing I'm worried about is that now this decoration is not usable in other places anymore, because it will cause unexpected behavior. Can you add a new property on the `MarginDecoration` class like `zeroMarginViewTypes` that is a list of integers? And then change this code here to check that the `viewType` is not in that list before applying the margin. That way, when using this class you can specify that the `0` and `1` viewTypes should not get a margin and get the same behavior as you have right now. While still having the current behavior of it applying a margin for all items.
A potential optimization for this would then also be to check if `zeroMarginViewTypes` is null or empty and if yes, skip the whole viewType calculation and apply the margin directly.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/inventory/StableRecyclerAdapter.kt
```diff
@@ -25,6 +26,9 @@ class StableRecyclerAdapter : androidx.recyclerview.widget.RecyclerView.Adapter<
     var itemType: String? = null
     var context: Context? = null
     var activity: MainActivity? = null
+
+    var shopSpriteSuffix: String = ""
```
phillipthelen: This is never set to anything, right? Or am I overlooking something?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/inventory/StableRecyclerAdapter.kt
```diff
@@ -52,15 +62,33 @@ class StableRecyclerAdapter : androidx.recyclerview.widget.RecyclerView.Adapter<
...
+    internal class StableHeaderViewHolder(itemView: View) : androidx.recyclerview.widget.RecyclerView.ViewHolder(itemView) {
+
+        private val npcBannerView: NPCBannerView by bindView(itemView, R.id.npcBannerView)
+        private val namePlate: TextView by bindView(itemView, R.id.namePlate)
+
+        fun bind(shopSpriteSuffix: String) {
+            npcBannerView.shopSpriteSuffix = shopSpriteSuffix
+            npcBannerView.identifier = "stable"
+
+            @Suppress("DEPRECATION")
```
phillipthelen: What deprecation is this suppressing here?

Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/inventory/StableRecyclerAdapter.kt
lelandhwu: Correct, this is just to keep it consistent with what was happening for shop headers, but it is not being used in this case

PR #1279, opened 2020-03-26
Android Multiplatform Integration

TNychka:
Creates Kotlin Multiplatform Shared Library for realm data models and adds local scoring interactor for offline mode.

This PR migrates the Android Realm models into a Shared Realm library. This Kotlin Mutliplatform development will enable the development of offline functionality that is mirrored and consistent across both iOS and Android applications. From the previous version of the PR, many models were condensed into a single form in only the shared instance. Only items that have clear dependencies on native platforms have been split into expected and actual classes, increasing the file count only slightly.


shared/build.gradle
```diff
@@ -56,6 +70,12 @@ kotlin {
...
 
+dependencies {
+    kapt 'io.realm:realm-annotations-processor:5.14.0'
+
+    implementation 'com.squareup.retrofit2:converter-gson:2.5.0'
+}
+
 // workaround for https://youtrack.jetbrains.com/issue/KT-27170
 configurations {
     compileClasspath
```
TNychka: If someone more experienced with Gradle could look at this closely I would appreciate it!

PR #1291, opened 2020-04-05
Fix unit tests

lelandhwu:
Got as many tests as I could running. 

There is a problem in which Roboelectric does not support Android Key Store, which is why some test suites are failing. Resources online say this is very difficult to fix, as some teams have actually forgone Roboelectric because of this. I think it's possible to fix, but will require some bit of time: https://github.com/robolectric/robolectric/issues/1518


.travis.yml
```diff
@@ -55,11 +58,11 @@ before_script:
   - cp habitica.resources.example habitica.resources
   - cp Habitica/google-services.json.example Habitica/google-services.json
 script:
-  - ./gradlew $TEST
+  - ./gradlew $TEST1
```
phillipthelen, line 61: If we can currently only run one of the tests, I would prefer to just run the `testProdDebugUnitTest` one. Since the `assembleDebug` does not actually run any tests but just confirms that the app can build a valid apk. (which is also useful, but having tests running is a bit more useful I think)

ryanmounla:
hey guys, any updates on that Key Store issue?

Issue #1045, opened 2018-09-11
Party invite: show sender in the UI


saraolson:
When you receive an invite and see it on the party screen, show the name/avatar of the user who sent the invite like this:
<img width="383" alt="screen shot 2018-09-11 at 10 32 08 am" src="https://user-images.githubusercontent.com/12453119/45366894-f9d11900-b5ad-11e8-9d5a-bd03f46e8f45.png">

ashleyng:
for the `accept` and `reject` assets, do i just go find some free ones that match as close as possible? It didn't look like there was anything similar already in the repo

saraolson:
@ashleyng so sorry for the delay, this slipped under our radar for a long time. if you're interested in picking this up again, i've included the icons in this folder.

[Party Invite Icons.zip](https://github.com/HabitRPG/habitica-android/files/2890036/Party.Invite.Icons.zip)

HiroinaProtagonist:
I'm interested in working on this. @ashleyng were you going to pick this back up?

ashleyng:
not at this moment, you can go ahead and take this if you'd like @HiroinaProtagonist

saraolson:
@HiroinaProtagonist i'm taking the help wanted label off and adding the in progress label for ya. please let us know if anything changes and you're not able to work on this, good luck!

saraolson:
@HiroinaProtagonist hi there, just wanted to check in to see how this is going? are you still working on it or do you have any questions you may need help with?

HiroinaProtagonist:
Hi, I'm still working on it. I don't have questions yet.

HiroinaProtagonist:
Just an update, I am working on this actively now and should have an update next week, if not sooner.

HiroinaProtagonist:
I do have a question: I have experience in several languages (it's not a difficult change) and I can partially localize the modified string. Would you like that done or what is the workflow for localization? (I'm happy to read some documentation if that already exists and I just haven't found it.)

Alys:
@HiroinaProtagonist All translations are done through Weblate rather than in GitHub. If you'd like to help with translations for any languages that you're a native speaker of, you'd be very welcome to join those translation teams, but it wouldn't be done through a pull request. The wiki's [Guidance for Linguists](https://habitica.fandom.com/wiki/Guidance_for_Linguists) has some details although Habitica has fairly recently moved from Transifex to Weblate and the new translation process is still being developed so some details on that page will be out of date.

[...]

PR #1297, opened 2020-04-11
WIP: Party invite

HiroinaProtagonist:
To address #1045 and other party-related issues that came up in testing.

Shows sender in the ui - separates front- and back-end changes
Removes invite view after a user rejects invite when not in a party
Redirects user to no party navigation after user leaves a party
Corrects a minor spelling error

I don't fully understand compositeSubscription and reactivex in general and will be rechecking that part and retesting. Any feedback is welcome.

my Habitica User-ID: HiroinaProtagonist


HiroinaProtagonist:
See updated pull request instead.

Issue #180, opened 2015-12-11
Make app compatible with Voice Over


lemoness:

AlgoRyan:
This is something that I have some experience with. Are there any particular requirements? How many users want this?

lemoness:
This has been requested on several occasions. As for the requirements, perhaps @viirus has some thoughts.

anitawoodruff:
I am happy to take a look at this. Stuff I plan to check:
- content descriptions - sensible labels are read out for UI elements when users explore by touch with Talkback. Purely decorative items are not read out at all.
- content grouping - adjacent content is read out together where it makes sense, but not when it doesn't.
- focus order - the right elements are focusable (purely decorative elements do not need to be focusable), so they are read out in a sensible order when the user swipes through them with Talkback enabled.
- live regions - changes to view triggered by user actions on other elements are announced, politely

Let me know if there's anything I'm missing!

phillipthelen:
@anitawoodruff That sounds good to me!

PR #1311, opened 2020-04-21
Talkback improvements for New Habit screen

anitawoodruff:
Third set of improvements towards making the whole app work well with Talkback - issue #180

- Previously when Talkback was enabled, it would read out all the button labels, but not whether they were currently selected - this was only represented visually by the colour of the buttons.

- Now, Talkback also reads out the current state (on/off, selected/not selected) of the following buttons:
  - Positive / negative habit buttons
  - Task difficulty buttons
  - Reset streak buttons

- This involved some code duplication, especially between the task difficulty buttons & reset streak buttons code. However, the logic being duplicated is fairly simple, so I think this is acceptable for now (especially given the classes already contain a fair amount of duplication.. but perhaps a larger refactor to eliminate this is due in future?).

- I also added `android:accessibilityHeading="true"` to the style of all headings in activity_task_view.xml - as advised at https://developer.android.com/guide/topics/ui/accessibility/principles#headings_within_text - so these are now read out as headings by Talkback.

### DIFF
--- a/Habitica/res/values/strings.xml
+++ b/Habitica/res/values/strings.xml
@@ -88,6 +88,8 @@
     <string name="negative_habit_form">Negative</string>
     <string name="on">On</string>
     <string name="off">Off</string>
+    <string name="selected">Selected</string>
+    <string name="not_selected">Not selected</string>
     <string name="checklist">Checklist</string>
     <string name="reminders">Reminders</string>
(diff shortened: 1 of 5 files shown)


anitawoodruff:
Superseded by #1317 1317

PR #1326, opened 2020-05-17
Fix grammar mistake on empty Special items page

kristian-welsh:
When you have no special items the following grammatically incorrect message is displayed
You don't have any Special
This is a special (hehe) case because the other item types already work in a plural context.

This commit changes that message to the following
You don't have any Special items
Or the equivalent in the user's language.

This may not be correct for right to left languages, as I have placed a space between Special and items. I have no way to verify this and don't know how to solve it if it is an issue.

my Habitica User-ID: `@Andriak`

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/inventory/items/ItemRecyclerFragment.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/inventory/items/ItemRecyclerFragment.kt
@@ -25,6 +25,7 @@ import com.habitrpg.android.habitica.ui.helpers.*
 import com.habitrpg.android.habitica.ui.views.HabiticaSnackbar
 import com.habitrpg.android.habitica.ui.views.HabiticaSnackbar.Companion.showSnackbar
 import io.reactivex.functions.Consumer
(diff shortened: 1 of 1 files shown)


Alys:
I'm not Habitica staff so if the staff say something different than this, listen to them not me. :)

One option would be to create a new string for "Special items" and use that in place of the itemTypeText string when the "special" category is being shown. That would allow translators to use appropriate word order and other grammatical adjustments that might be needed.

But I'm also not an Android coder so I don't know how well this would fit into the current code. :)

kristian-welsh:
That's true. I chose this approach since it was the simplest thing I could do without translator support. Happy to make the code change if we can get translations for the full phrase "You don't have any Special items" I'm not sure how habitica manages translation effort.

Alys:
I don't think we'd want a full string for "You don't have any Special items". We just need one for "Special items" and then use that instead of the string for "Special" when you're viewing the "Special" screen.

You don't need to worry about translations. Just add the English string and then the translation process will handle the rest (if you're curious, [this wiki page](https://habitica.fandom.com/wiki/Guidance_for_Linguists) describes how translations are done).

kristian-welsh:
That should be suitable

Issue #1184, opened 2019-06-11
Widget: How to view only today's To-Do's in the widget on Android 


Franklin1997:
I want to see only today's to-do's in the android widget so that I can focus on tasks to do today and not all the tasks from the future. Is there any option available to see only today's to-do's?

PR #1353, opened 2020-08-21
Only todos for today#1184

PatriciaHeimfarth:
For https://github.com/HabitRPG/habitica-android/issues/1184
my Habitica User-ID: @PehHah


phillipthelen:
This looks like it adds a new filter option to the task list in the app, but not to the widget as mentioned in the issue?

PatriciaHeimfarth:
> This looks like it adds a new filter option to the task list in the app, but not to the widget as mentioned in the issue?

Yeah, I think you are right. I will do another change only for the widget. :)

Issue #1306, opened 2020-04-16
Quests: No indication of finding a collection quest item


saraolson:
There is no toast like web has of +1 Quest Item found, and there's no indication of pending items in the party UI

Quazyrog:
Can I try to fix it? I haven't touched Habitica code before, but I study programming and I had some experience with Android application programming (in Kotlin) recently.

saraolson:
@Quazyrog If you're still interested in this, we'd love to have you try to fix it! Thanks for your patience, things have been a little hectic recently! :)

Quazyrog:
I'm still interested! I finished setting up server for testing (maybe it will be useful somehow to have it locally), so I can start playing with the code this week :D

saraolson:
@Quazyrog perfect! ive marked the issue as in progress, let us know if you have any questions :)

Quazyrog:
So it seems to be working now [here](https://github.com/Quazyrog/habitica-android/tree/collection-quest-indication). I'm waiting yet to see if changes in World Boss screen haven't broken anything — I asked about a way to start World Boss with local Habitica installation so I should be able to see it soon.

I guess I should rebase before making PR?

phillipthelen:
When you create the PR in GitHub it should show you the changes that belong to the PR and if there is a merge conflict. As long as there is no conflict, it should be fine without rebasing.

Quazyrog:
There was a merge conflict in translations file xD

Also I tested the world boss screen and it seems to be working after my changes as far as I can tell. This is also the cause of this delay — sorry.

saraolson:
@Quazyrog thanks for the PR! we're in a rush to push out the 3.0 milestone so we'll be reviewing PRs and offering feedback during the 3.1 milestone

PR #1354, opened 2020-08-25
Collection quest indication

Quazyrog:
Number of found items shown when user is on collection quest (#1306)

1. After completing a task/habit/... the number of found quest items is shown in a toast together with other looted items.
2. Cumulative number of collected items since last cron is shown in party screen below progress bars.
3. Both of these should only show for collection quest (as long as Realm fetched enough data to identify this as a collection quest... it seems that viewing party screen with the quest at least once should do the trick). Similarly for damage.

my Habitica User-ID: 0896a7a9-8dfb-4112-81cc-518cb8d14c08


phillipthelen:
Thanks! This looks great. I'll note this as a contribution in your Habitica profile.

Issue #1302, opened 2020-04-15
Messages: Add initial message UI


saraolson:
We have designs for UI that should show when you're starting your first message with someone. It shows their display name, username, and a string about being nice. optionally: shows their current avatar. This should stay on the screen when the conversation starts, pushed up by the convo

<img width="394" alt="Screen Shot 2020-05-27 at 10 07 40 AM" src="https://user-images.githubusercontent.com/12453119/83030919-f25fd680-a001-11ea-8876-5f652009cd16.png">

laserzt:
Hi, I would like to start to work on this. I have a question, we don't want to change the "Send Message" dialog, right?
Only the messaging dialog. Even if someone sends a message using the "Send Message" to someone new, it still should open the old Dialog box, right?

saraolson:
@laserzt are you talking about the modal that pops up and prompts you to enter a username? there aren't any changes there. this UI will show after you enter a username and are shown the new conversation screen. You can see the initial screenshot in more context below:
<img width="288" alt="Screen Shot 2020-08-13 at 10 54 33 AM" src="https://user-images.githubusercontent.com/12453119/90150297-83b62d80-dd53-11ea-9637-f23d80fed577.png">

(there are some UI elements in this screenshot that don't match up with the current implementation ie the bottom bar, top bar, and 'Today' label, ignore those)

laserzt:
I'm talking about the dialog that open what you press "Send Message" in a person's profile:
![image](https://user-images.githubusercontent.com/1330791/90166875-8b052780-dda3-11ea-90c3-f63fa10f05b2.png)

saraolson:
@laserzt ah, okay. that must be really implementation. in this case, we actually would be getting rid of that flow. opting to send someone a message from their profile should just open the normal conversation screen you'd see if you went to messages.

laserzt:
Ok. So, I'm starting to work, ok?

saraolson:
@laserzt yup, ill mark it as in progress!

laserzt:
I made a pull request #1355

saraolson:
we're in a rush to push out the next milestone so we'll be reviewing PRs and offering feedback for milestone 3.1

PR #1355, opened 2020-08-26
issue #1302: "Messages: Add initial message" fix

laserzt:
my Habitica User-ID: 08576d02-d94b-48aa-9ed3-16de73dd814b

There is still a bug, that existed before my change, that a new user cannot be messaged by writing his name in the inbox, and if you press the home button, the program crashes. The bug is only in develop, and not in production.
I haven't changed the behavior of the Message dialog from the user profile. Sorry.... Tried, Couldn't do it.


phillipthelen:
Thank you! I notes this as a contribution in your Habitica profile.

PR #1356, opened 2020-08-27
todos widget has only tasks with todaydue or nulldate

PatriciaHeimfarth:
For https://github.com/HabitRPG/habitica-android/issues/1184
my Habitica User-ID: @PehHah


phillipthelen:
@saraolson What do you think? With this change, the widget would only display todos for today or without a due date. I'm not sure if we should rather have this as an optional setting or make this a separate widget.

saraolson:
@phillipthelen the original issue was status:needs discussion and not marked help wanted because we werent sure how or if we wanted the feature. i dont want to change the way the todo widget works right now or feel comfortable making this decision without taking the time to look into the widget, the use case, and our options further.

Issue #1303, opened 2020-04-15
Messages: swap order of messages so they match other chats


saraolson:
Newest message is at the top of the screen at the moment, should be at bottom to match chats. also add a hard refresh button to message screen

PatriciaHeimfarth:
I would like to work on this issue

saraolson:
@PatriciaHeimfarth Marked as in progress for you! We're in a rush to push out the next milestone so we'll be reviewing PRs and offering feedback during the 3.1 milestone

PR #1357, opened 2020-08-28
no descending sorting of inbox messages

PatriciaHeimfarth:
For https://github.com/HabitRPG/habitica-android/issues/1303
my Habitica User-ID: @PehHah


phillipthelen:
I don't think this will have the right UX. While the order would now be reversed, the recycler view still works from top to bottom. Meaning the app would open with the most oldest messages up top and the newest ones off screen at the bottom (given there are enough for the view to be scrolling)
This probably needs similar changes to what the ChatFragment has to reverse the recyclerview

phillipthelen:
@PatriciaHeimfarth Do you think you are able to complete this PR? Otherwise I'll pick this up soon, so that we can have it in 3.1

Issue #1319, opened 2020-05-02
To-Dos aren't filtered on startup


Gansgar:
**Describe the bug**
When I open the app, I first see the to-do tab. Which is no problem, but the list contains even all check-off tasks. 

When I try to filter the list it says that the list is filtered with active only. I need to switch the filter to another one and switch back to the "actives only" filter to actually just see my active tasks.

**To Reproduce**
I haven't found a concise way yet to reproduce the issue. When I stop the app completly and open it again it works. I.e. when I try to force the behaviour. 
But this bug occurs to me every single time I opened the app the past few days. 

EDIt: It seems that when the app is in the background and isn't touched for a few hours (probably removed from RAM), reopening it produces the error. Restarting the app by closing it, fixes the error again.

**Expected behavior**
The To-Do list should be filtered with active tasks only.

**Smartphone (please complete the following information):**
 - Device: Samsung S10
 - OS: Android 10
 - App Version: 2.6.1
 - Habitica Username: gansga

PatriciaHeimfarth:
I would like to work on that issue

shane-tang:
👋 If it's okay I'll be picking this up for my first contribution. I took a quick look last night and I think I found the solution

saraolson:
@shane-tang thanks for picking this up! we're working on our iOS release right now but we'll review this for android's next milestone.

PR #1363, opened 2020-09-24
Fix default todo filters on boot

shane-tang:
This fixes Issue #1319.

Steps to reproduce bugged behavior:
1. Open Habitica
2. Complete a Todo task.
3. Close the app (important: while still on the todo fragment)
4. Set the background process limit to 0
    * Developer options -> Background process limit -> "no background processes"
5. Open another app
    * From my understanding, this will remove as much app state from the device memory as possible without actually force stopping the app
6. Open Habitica
    * The app will seem like it's cold booting, but it's actually going to restore into the todo fragment instead of the habit fragment

My Habitica User-ID: 99f20106-7060-4398-b565-26e56c4de2ad

| Bugged Behavior | Fixed Behavior |
| ----------------- | --------------- |
| 1319 repro | 1319 fixed |

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.kt
@@ -163,12 +163,15 @@ open class TaskRecyclerViewFragment : BaseFragment<FragmentRefreshRecyclerviewBi
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         context?.let { binding?.recyclerView?.setBackgroundColor(ContextCompat.getColor(it, R.color.content_background)) }
-        if (Task.TYPE_DAILY == classType) {
-            if (user?.isValid == true && user?.preferences?.dailyDueDefaultView == true) {
-                taskFilterHelper.setActiveFilter(Task.TYPE_DAILY, Task.FILTER_ACTIVE)
+        savedInstanceState?.let { this.classType = savedInstanceState.getString(CLASS_TYPE_KEY, "") }
+
+        when (classType) {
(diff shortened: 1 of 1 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.kt
```diff
@@ -163,12 +163,15 @@ open class TaskRecyclerViewFragment : BaseFragment<FragmentRefreshRecyclerviewBi
...
-            if (user?.isValid == true && user?.preferences?.dailyDueDefaultView == true) {
-                taskFilterHelper.setActiveFilter(Task.TYPE_DAILY, Task.FILTER_ACTIVE)
+        savedInstanceState?.let { this.classType = savedInstanceState.getString(CLASS_TYPE_KEY, "") }
+
+        when (classType) {
+            Task.TYPE_TODO -> taskFilterHelper.setActiveFilter(Task.TYPE_TODO, Task.FILTER_ACTIVE)
+            Task.TYPE_DAILY -> {
+                if (user?.isValid == true && user?.preferences?.dailyDueDefaultView == true) {
+                    taskFilterHelper.setActiveFilter(Task.TYPE_DAILY, Task.FILTER_ACTIVE)
+                }
```
shane-tang, line 173: We check the `classType` to decide which filters to apply. The issue was `classType` was being assigned *after* the filters are applied, so from a cold boot we'll be checking against a `null` `classType`. This hasn't been noticed before because *most* of the time a cold boot will load into the habits tab, whose default filter is "all."

shane-tang: Also Kotlin-ized some things for readability

phillipthelen:
Thank you for this find! I noted this as a contribution in your Habitica profile.

Issue #1427, opened 2020-11-27
Setting language to Hebrew creates odd appearance of settings page


shanaqui:
![_20201126_090549](https://user-images.githubusercontent.com/8144640/100445861-b3d8d080-30a5-11eb-982c-4754e140783f.JPG)

Device: Sony F5321
Android Version: 25
AppVersion: Version 3.1 (2660)
BETA
User ID: 33bb14bd-814d-40cb-98a4-7b76a752761c
Level: 81
Class: rogue
Is in Inn: false
Uses Costume: true
Custom Day Start: 16
Timezone Offset: -60

saraolson:
is the desired fix just that the settings page has a longer line length? im guessing the right side text is becuase hebrew is a rtl language, but we dont have translations in place for settings?

Nakonana:
@saraolson yeah looks like the heading ("Profile" etc.) is defining the line length for its notes.

As for the translations, I just checked and the majority of settings strings is indeed not translated, but there's definitely also a translation bug present as can be seen by the German word "Einstellungen" in the screenshot above but also by the fact that at least for "Logout" there's a Hebrew translation:

![Screenshot_20201130-190701](https://user-images.githubusercontent.com/68129995/100647547-fc131f80-333f-11eb-8808-ebc51883d048.png)

saraolson:
@Nakonana thanks for the clarification!

JofielB:
I was able to reproduce the error and I think that I am starting to know where is the problem. Let me explain the experiment that I made 
The first thing that I noticed is that when you change the language from English to Hebrew some UI elements from the application are mirrored (I still don't know if this is a feature of the app or if it is something that android does for some languages). And I believe that this is the root of the problem because I was using two apps of Habitica, the one from the play store and the other one is from the project and when I changed the languages to Hebrew from the project one the UI from the app of the play store was mirrored to and the error occurred too but the language was in English. 
So I am pretty sure that the problem is because of that flip of the UI that happens when the language is changed

### Normal UI
![Issue 2](https://user-images.githubusercontent.com/32725639/100810646-22939200-33f6-11eb-876a-0330ee34c6fc.jpg)

### Mirrored UI (this happened when you change the languages)
![Issue 3](https://user-images.githubusercontent.com/32725639/100810876-9df54380-33f6-11eb-9d5b-c805b8d32bbd.jpg)

### Error reproduced with the English language because of the mirrored UI
[...]

PR #1431, opened 2020-12-03
Fix odd appearance on Settings page when using an RTL(Right to left) languages

JofielB:
### Summary
This pull request is to resolve issue #1427. The problem was in the alignment properties of the summary TextView, with the old properties of alignment the TextView moved in a no desire way when the app was mirrored for an RTL language
### Changes
- Modify `preference_child_summary.xml`
### Target
`develop`

my Habitica User-ID: 78b5d1f8-b295-4f95-9541-3d1b418358c0


saraolson:
we'll review PRs and merge them when we're working on the next android update! 3.1 just came out so we're working on iOS right now, meaning it might be a bit of a wait til we get back around to this but it won't be forgotten :)

Issue #1430, opened 2020-12-02
Appears to be possible to "like" your own posts


shanaqui:
User appears to be able to like their own posts, instead of receiving the error. If they close the app and come back, though, the self-like is gone.

![Screenshot_20201201-202518](https://user-images.githubusercontent.com/8144640/100866841-27595400-3491-11eb-98ed-880bbc982467.png)

Device: Sony F5321
Android Version: 25
AppVersion: Version 3.1 (2663)
BETA
User ID: 33bb14bd-814d-40cb-98a4-7b76a752761c
Level: 81
Class: rogue
Is in Inn: false
Uses Costume: true
Custom Day Start: 16
Timezone Offset: -60

JofielB:
Hello, This issue is available? I would like to take it

saraolson:
@JofielB it is available, and now its yours :) good luck!

JofielB:
On that pull request, an alert is shown when you try to like your own post.
![PR](https://user-images.githubusercontent.com/32725639/101432486-2b88d580-38c6-11eb-9165-938a7941217b.png)

PR #1435, opened 2020-12-08
Show an alert when a user try to like its own posts

JofielB:
### Summary
This pull request is to resolve issue #1430. 
When a user likes its own posts an alert is shown and the +1 in hist post doesn't appear.

### Changes
Modify `ChatRecyclerMessageViewHolder`

### Target
`develop`

my Habitica User-ID: 78b5d1f8-b295-4f95-9541-3d1b418358c0


saraolson:
could we change this from a modal you have to dismiss to a toast that pops up at the bottom of the screen and dismisses itself?

Issue #1460, opened 2021-01-15
Phantom pet eggs


shanaqui:
Users have been reporting this on and off for a while.

> "Hi, so I've had this glitch in my egg inventory for probably at least 9-12 months.
> 
> I think I used to know what the 12 missing items were, but I forgot ages ago.
> 
> If I tap on them, the pop up for selling an egg shows up and offers me 'null' gold. I haven't tried to sell any for null gold, at least not in the last few months or so.
> 
> It's not a huge problem, especially since I can't remember what they were anymore, but it is odd and it is a glitch and it hasn't ever gone away that I've caught."
> 
> Device: samsung SM-A515U
> Android Version: 29
> AppVersion: Version 3.1.1 (2690)
> User ID: 31ff377b-690f-4328-952e-0aab8fb8e80e
> Level: 134
> Class: wizard
> Is in Inn: false
> Uses Costume: true
> Custom Day Start: 3
> Timezone Offset: 300

They don't have twelve of anything in their inventory.

![Screenshot_20210114-020745_Habitica](https://user-images.githubusercontent.com/8144640/104735016-fb401c80-5738-11eb-804e-7c473fcccb55.jpg)

saraolson:
I also have this problem, and have brought it up but we've never found a cause or a fix for them. They dont seem to be actual items, just some bad data? ie nothing is missing that the account should have access to

shanaqui:
Also currently being experienced by this user:

Device: samsung Galaxy S8
Android Version: 28
AppVersion: Version 3.1.2 (2694)
User ID: 243484ac-bb24-4170-8fe7-bf3a53e0684b
Level: 41
Class: wizard
Is in Inn: false
Uses Costume: false
Custom Day Start: 7
Timezone Offset: 0

shanaqui:
And with an interesting new detail:

> Device: samsung SM-J600FN
> Android Version: 26
> AppVersion: Version 3.2 (2832)
> User ID: b24c2422-7212-448f-a5bb-c9603d25c861
> Level: 7
> Class: warrior
> Is in Inn: false
> Uses Costume: false
> Custom Day Start: 0
> Timezone Offset: 0
> Details:
> 
>  Please see screenshot for the inventory_present item in my Egg Inventory. It only gives the option to Sell for null Gold. 

Maybe something weird with the subscriber giftbox?

[...]

PR #1526, opened 2021-04-05
1460 phantom pet eggs

cvuorinen:
This fixes the issue of eggs with title `inventory_present` show in Items > Eggs (from issue #1460). Found out that they are actually the Mystery items that get mistakenly categorized as eggs when updating user data.

That change in `SpecialItems.kt` is the one that actually fixes the problem, but I made that other change too since it felt like appropriate (and I first landed there and thought it was the cause of the issue) as I thought the list should match this other one: https://github.com/HabitRPG/habitica-android/blob/ff6dac2aae34f6a9a07fbf1420ad11e8d2c0ee28/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/inventory/items/ItemRecyclerFragment.kt#L219-L226

The problem was not shown initially when you go to the Items view, but appeared sometime later. I found out that setter method in SpecialItems is called from https://github.com/HabitRPG/habitica-android/blob/ff6dac2aae34f6a9a07fbf1420ad11e8d2c0ee28/Habitica/src/main/java/com/habitrpg/android/habitica/models/user/Items.kt#L18 which is called from https://github.com/HabitRPG/habitica-android/blob/ff6dac2aae34f6a9a07fbf1420ad11e8d2c0ee28/Habitica/src/main/java/com/habitrpg/android/habitica/models/user/User.kt#L112 which is called by `mergeUser` in https://github.com/HabitRPG/habitica-android/blob/ff6dac2aae34f6a9a07fbf1420ad11e8d2c0ee28/Habitica/src/main/java/com/habitrpg/android/habitica/data/implementation/UserRepositoryImpl.kt#L412 which happens when you go the Settings view.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/data/local/implementation/RealmInventoryLocalRepository.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/data/local/implementation/RealmInventoryLocalRepository.kt
@@ -229,6 +229,7 @@ class RealmInventoryLocalRepository(realm: Realm) : RealmContentLocalRepository(
             "hatchingPotions" -> HatchingPotion::class.java
             "food" -> Food::class.java
             "quests" -> QuestContent::class.java
+            "special" -> SpecialItem::class.java
             else -> Egg::class.java
         }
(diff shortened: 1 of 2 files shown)


phillipthelen:
Thanks! This looks like it does fix the issue.

Issue #1540, opened 2021-05-03
Screen reading software cannot navigate to completion focus


qq3477906169:
Hello developer, I am a visually impaired user, I cannot complete Dailies because the screen reader cannot navigate to the completed focus, I hope to solve it, thank you very much
Device: HUAWEI GLK-AL00
Android Version: 29
AppVersion: 版本3.2.3 (2934)
User ID: 88079191-691e-4d5f-9ed5-b7b12611e3ab
Level: 5
Class: rogue
Is in Inn: false
Uses Costume: true
Custom Day Start: 0
Timezone Offset: -480
Details:

Trion129:
I'll take a look.

Trion129:
Tried adding contentDescription to the button. Then installed Talkback app to see the experience. Not able to focus on the specific checkbox yet. Will try to fix it. While at it I also found that Todo has similar problem. Will try to fix it too.

Trion129:
Got it to focus on the checkbox and speak it out. But I found another bug with this. So, when I use talkback to click any of the buttons, it doesn't update the state. This is for all buttons. So, let's say I press + in Habits, it doesn't update the streak count. A checkbox in dailies doesn't mark the object as completed. Will try to fix this too, interesting to learn about accessibility through this issue.

PR #1570, opened 2021-06-17
Fixed accessibility focus on Daily checkbox and todo checkbox. fixed transaction ran on button click.

Trion129:
Addresses #1540 

The user was not able to select the Daily checkbox. I made it focusable and added a content description.
I found the same issue was happening in the Todo list, so fixed that too.

While I was working on this fix, I noticed that when I click any task buttons, it caused an error from Realm that Object is not available. This caused DB to update, but UI got out of sync.
I read the code and found that we are storing an object by reference and there is a section in Realm that deletes this object, replacing it with a new Proxy object. That's why the object was becoming unavailable.
So I replaced the reference with a copyFromRealm, which will create a copy of the variable instead of a mutable reference.

My Habitica User-ID: _trion

### DIFF
--- a/Habitica/res/layout/daily_item_card.xml
+++ b/Habitica/res/layout/daily_item_card.xml
@@ -26,7 +26,9 @@
                     android:id="@+id/checkBoxHolder"
                     android:layout_width="@dimen/button_width"
                     android:layout_height="match_parent"
-                    tools:background="@color/red_10">
+                    tools:background="@color/red_10"
+                    android:focusable="true"
+                    android:contentDescription="@string/daily_item_checkbox">
                     <View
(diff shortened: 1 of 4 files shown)


Trion129:
Cannot add reviewers from UI, so mentioning. @phillipthelen please review the PR when you have time.

saraolson:
thanks for your contribution! will try to merge this into 3.3, but since 3.3 is so close to release, if there's any issues, it may get pushed to 3.4.

Trion129:
Hi @saraolson @phillipthelen Can someone also credit this PR towards Contributor tiers?

phillipthelen:
Sorry for the long wait. I attributed this contribution to you.

PR #1662, opened 2021-10-23
Bug fix message notification not opening chat

### LINKED ISSUE #1617: DM push notif doesn't open app to the DM
Ideally we would load the actually conversation, but right now the app doesn't even take you to messages, just tasks.

pauliancu97:
my Habitica User-ID: paul97iancu
Closes #1617
I added a fix for the problem of the user not being redirected to the correct conversions when the message notification is tapped. However, while sending messages using Firebase Messaging and sending Display Messages, I noticed that the notification built by our app was shown only if the app was already in the foreground, if the app was in the background, the default notification was shown. If instead of Display Messages I sent Data Messages, the app built notification was shown even if the app was in background. So I guess that maybe there is also a backend problem where instead of Data Messages the backend sends Display Messages.


saraolson:
@pauliancu97 i've added this to your habitica profile and promoted you to contributor tier 3 🎊 thanks for all your help with the android app!

PR #1666, opened 2021-11-06
Hide special pets mounts if empty

### LINKED ISSUE #1647: Hide 'Special Mounts/Special Pets' category when there are none
Right now when you don't have any special mounts, it looks like the app isn't scrolling properly since we don't show placeholder silhouettes for these mounts that may never be available again. Probably best to just hide the category label all together when nothing is there.

pauliancu97:
my Habitica User-ID:paul97iancu

Closes #1647


saraolson:
@pauliancu97 thanks for your contribution! i've recorded it on your habitica profile

PR #1667, opened 2021-11-06
Disable changes for tasks in challenges

### LINKED ISSUE #1588: Appears to be allowing changes to challenge tasks which then are not saved
> Device: Xiaomi Mi MIX 2
> Android Version: 28
> AppVersion: Version 3.2.4.2 (2966)
> User ID: 35c3fb6f-fb98-4bc3-b57a-ac01137d0847
> Level: 3
> Class: warrior
> Is in Inn: false
> Uses Costume: false
> Custom Day Start: 7
> Timezone Offset: 420
> Details:
> 
>  Hi,
> 
> If I try to edit the name of a challenge task, the app allows me to edit and save it, but then when I check the name of the task, it hasn't actually changed. This misleads the user into thinking they have edited the name, but if they don't double-check, they won't notice that their edit wasn't saved! The same is true for any field that cannot be changed by the user, according to this wiki page: https://habitica.fandom.com/wiki/Challenges

pauliancu97:
my Habitica User-ID:paul97iancu
Closes #1588 
I saw that the web version disables the uneditable fields in a task that belongs to a challenge.


saraolson:
thanks again! added this to your habitica profile as well

PR #1704, opened 2022-02-04
Discard Modal for user editing a Task and presses back

Hafizzle:
Discard Modal prompts user to Discard or cancel edited task - For both current & new tasks.
Compare task in progress to New/Current task
Firing events if user either presses Discard, or cancels discard (or dismisses dialog) and saves task.
Event: discard_task Bundle: boolean is_creating
Event: back_to_task Bundle: boolean is_creating


saraolson:
Added this to your habitica profile and granted you contributor rank 1 :D

Issue #1349, opened 2020-07-30
Show transformation items in Items>Special and allow use from there


saraolson:
When you buy a transformation item, you expect it to show in the items section (as it does on web). By showing it only in the skills section, it locks people out that may not have skills yet (low level, opted out of classes). 

Flow:
Tap item > bottom menu comes up with option 'Use' > show screen that displays party members
Basically copy the same flow as using the item from Skills, but by tapping the item instead.

I think showing the same item in two places is fine for now, as it is a bit more convenient to use from skills.

Mike-Antonacci:
Is the desired behavior to keep the transformation items under Skills and additionally add them to Items, or to only have them in Items and eliminate finding them in Skills?

Should the item appear under Inventory > Items > Special (from the sidebar), or be pinned in Rewards, or both?

Does this apply to just the initial transformational item (seafoam, etc.) or also the counter-transformational item to undo the transformation?

saraolson:
@Mike-Antonacci 
1. Yes, desired behavior is to still allow use from Skills, just adding the ability to also use from Items.

2. Yes, the location would be Inventory > Items > Special. Pinning to Rewards is for buying the item, and users are always allowed to pin any item they can currently buy to Rewards. If they were to pin Seafoam, then what should be shown in Rewards is the UI to buy the item. If they then bought Seafoam from Rewards, the item would be stocked in their inventory and they could use it whenever they want. Initial transformation items aren't instant use.

3. This only applies to the initial transformation item, as the counter-transformation items are instant-use and are never added to your inventory, similar to a health potion.

hope this helps clear things up.

Mike-Antonacci:
then in that case, I'd love to work on this one.

saraolson:
@Mike-Antonacci awesome, ill mark it as in progress for you!

Mike-Antonacci:
Is there an Android equivalent of setting up a local Habitica install for testing purposes, like Blacksmiths do for the web? I don't want to make changes without being able to see what the changes are

Mike-Antonacci:
As I started work on this one, I realized it's beyond my current knowledge and was more difficult than I expected. I'd like to stop working on it and allow someone else to do it instead.

PR #1706, opened 2022-02-14
#1349 Issue - Add ability to use transformation items from Items with same functionality as from Skills

Hafizzle:
- You will now be able to use transformation items from Items as well with the same functionality as from the Skills Tab
- Includes transformation items but not counter transformation items


Habitica/src/main/java/com/habitrpg/android/habitica/data/local/implementation/RealmUserLocalRepository.kt
```diff
@@ -147,16 +147,16 @@ class RealmUserLocalRepository(realm: Realm) : RealmBaseLocalRepository(realm),
...
-            if (specialItems.shinySeed > 0) {
+            if (specialItems.firstOrNull() { it.key == "shinySeed" }?.numberOwned ?: 0 > 0) {
                 ownedItems.add("shinySeed")
             }
-            if (specialItems.seafoam > 0) {
+            if (specialItems.firstOrNull() { it.key == "seafoam" }?.numberOwned ?: 0 > 0) {
                 ownedItems.add("seafoam")
             }
-            if (specialItems.spookySparkles > 0) {
+            if (specialItems.firstOrNull() { it.key == "spookySparkles" }?.numberOwned ?: 0 > 0) {
```
phillipthelen, line 159: Can you replace this with a for loop? Something like this:
```
for (key in listOf("snowball", "shinySeed", "seafoam", "spookySparkles")) {
    If (specialItems.firstOrNull() { it.key == key }?.numberOwned ?: 0 > 0) {
        ownedItems.add(key)
    }
}
```

Just to remove some code duplication and make adding new keys easier.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/adapter/SkillsRecyclerViewAdapter.kt
```diff
@@ -135,10 +136,10 @@ class SkillsRecyclerViewAdapter : RecyclerView.Adapter<SkillsRecyclerViewAdapter
...
         private fun getOwnedCount(key: String): Int {
             return when (key) {
-                "snowball" -> specialItems?.snowball
-                "shinySeed" -> specialItems?.shinySeed
-                "seafoam" -> specialItems?.seafoam
-                "spookySparkles" -> specialItems?.spookySparkles
+                "snowball" -> specialItems?.firstOrNull() { it.key == "snowball" }?.numberOwned
+                "shinySeed" -> specialItems?.firstOrNull() { it.key == "shinySeed" }?.numberOwned
+                "seafoam" -> specialItems?.firstOrNull() { it.key == "seafoam" }?.numberOwned
+                "spookySparkles" -> specialItems?.firstOrNull() { it.key == "spookySparkles" }?.numberOwned
```
phillipthelen, line 142: replace the whole `when` part with just the check. 

```
return specialItems?.firstOrNull() { it.key == key  }?.numberOwned ?: 0
```

this would do the same thing and also handle new keys without needing changes.

Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/NavigationDrawerFragment.kt
```diff
@@ -342,7 +342,7 @@ class NavigationDrawerFragment : DialogFragment() {
         val specialItems = user.items?.special
         var hasSpecialItems = false
         if (specialItems != null) {
-            hasSpecialItems = specialItems.hasSpecialItems
+            hasSpecialItems = true
```
phillipthelen, line 345: This would show the skills menu item in some cases where we don't want it. I think instead the `Items` class should have a method to check if the user has transformation items (with the same logic from the previous `hasSpecialItems` method. To clarify the usage that should then be called `hasTransformationItems` instead, since it wouldn't count cards or mystery gifts

Habitica/src/main/java/com/habitrpg/android/habitica/utils/ContentDeserializer.kt
```diff
@@ -101,6 +94,20 @@ class ContentDeserializer : JsonDeserializer<ContentResult> {
                 result.spells.add(skill)
             }
         }
+
+        if (obj.has("special")) {
+            for ((classname, value) in obj.getAsJsonObject("special").entrySet()) {
+                val skillItem = value.asJsonObject
```
phillipthelen, line 100: Is there a reason against using `context.deserialize(obj.get("special"), object : TypeToken<RealmList<SpecialItem>>() {}.type)` like we do for other entries?

phillipthelen (commented):
Left some thoughts, but overall looks good so far!

PR #1717, opened 2022-02-25
Show category of item in 'You dont have any item' error instead of null.

### LINKED ISSUE #1663: 'You don't have any null' shows when using an item with nothing to use it on
to reproduce, go to items and try to use a hatching potion with no eggs. It should say 'You don't have any Eggs' or if the other way around, 'You don't have any Hatching Potions' 
Screenshot_20211025-150653
s.

dhruvnagarajan:
Closes #1663

my Habitica User-ID: quazared


saraolson:
@dhruvnagarajan thanks for your contribution! I've credited it to your Habitica profile and awarded you a contributor tier 🎊

PR #1720, opened 2022-03-01
Fixes#1258 - Fixed reminders occurring randomly or not at all

Hafizzle:
Replace setAndAllowWhileIdle with setAlarmClock due to both setAndAllowWhileIdle and setExactAndAllowWhileIdle have triggering alarm restrictions - setAlarmClock is exact and does not limit alarm count.


Hafizzle:
Used following flowchart to decide -
image

Tested across multiple reminders successfully.

PR #1747, opened 2022-04-11
Fixes#1637 - Fix users unable to purchase gems because "you already own this item"

Hafizzle:
queryPurchasesAsync was moved from init block to onResume (per Android billing documentation).

onQueryPurchasesResponse was not returning purchases that needed to be consumed if onPurchasesUpdated did not catch them due to queryPurchasesAsync being run on init block.


Hafizzle:
@phillipthelen I'm going to need help testing this but I believe this miggght explain why any consumable that slips through the cracks (not being consumed onPurchasesUpdated) aren't being caught by onQueryPurchasesUpdated - queryPurchasesAsync wasn't being called onResume.

image

PR #1752, opened 2022-04-18
Fixes#1678 - Alarms support timezone changes

Hafizzle:
Implement ZonedDateTime to handle timezone changes for reminders / alarms


Hafizzle:
Tested timezone changes including with repeat alarms successfully.

PR #1760, opened 2022-04-29
Fixes#1347 - Fix line below unfocused text boxes to show correctly

Hafizzle:
Line shows correctly & colors updated

(purple 50, yellow is yellow 5, the rest are color 10)


phillipthelen:
This seems to have the same issue as the other one with unrelated changes being included

PR #1761, opened 2022-04-29
Fixes#1716 - Fix party launch not working reliably

Hafizzle:
Await observer (once) to check if user is currently has party set as launch, and navigate to Party screen.


phillipthelen:
It looks like this includes some changes that are unrelated to the problem like changes to the task form, purchase handler and items. It also removes the two example config files and downgrades the Gradle version. Can you take a look at that?

But the actual fix for this issue looks good!

PR #1764, opened 2022-05-02
Fixes#1722 - Show next hourglass month on sub info screen

Hafizzle:
Set when user will next get hourglass as month using offset (Using Java.Time)

### DIFF
--- a/Habitica/res/layout/subscription_details.xml
+++ b/Habitica/res/layout/subscription_details.xml
@@ -183,13 +183,13 @@
                 android:layout_width="0dp"
                 android:layout_weight="1"
                 android:layout_height="wrap_content"
-                android:text="@string/current_hourclasses"
+                android:text="@string/next_hourglass"
                 style="@style/subscriptionBoxText.Subtitle" />
             <TextView
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 style="@style/Pill"
-                android:id="@+id/currentHourglassesTextView"
+                android:id="@+id/nextHourglassTextView"
                 tools:text="4" />
         </LinearLayout>
     </LinearLayout>
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/subscriptions/SubscriptionDetailsView.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/subscriptions/SubscriptionDetailsView.kt
@@ -4,6 +4,8 @@ import android.content.Context
 import android.content.Intent
 import android.graphics.drawable.BitmapDrawable
 import android.net.Uri
+import android.text.format.DateUtils
+import android.text.format.DateUtils.getRelativeTimeSpanString
 import android.util.AttributeSet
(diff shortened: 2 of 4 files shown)


Hafizzle:
@phillipthelen This PR now handles when a basic subscription user will receive their next hourglass (On the fourth month)

Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/subscriptions/SubscriptionDetailsView.kt
```diff
@@ -96,8 +105,15 @@ class SubscriptionDetailsView : LinearLayout {
         } else {
             binding.monthsSubscribedTextView.text = resources.getString(R.string.x_months, plan.consecutive?.count ?: 0)
         }
+
         binding.gemCapTextView.text = plan.totalNumberOfGems.toString()
-        binding.currentHourglassesTextView.text = plan.consecutive?.trinkets.toString()
+
+        if (plan.consecutive?.count != null) {
```
phillipthelen, line 111: Can you move this calculation out of the view and into either a separate class or the SubscriptionPlan class? And then add a unit test to it, so that we can make it check a few cases and be sure it works as expected.

Habitica/res/layout/subscription_details.xml
```diff
@@ -183,13 +183,13 @@
...
                 android:layout_height="wrap_content"
-                android:text="@string/current_hourclasses"
+                android:text="@string/next_hourglass"
                 style="@style/subscriptionBoxText.Subtitle" />
             <TextView
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 style="@style/Pill"
-                android:id="@+id/currentHourglassesTextView"
+                android:id="@+id/nextHourglassTextView"
```
phillipthelen, line 192: When creating/renaming I started to switch to using snake_case instead of camelCase in the layout file, because the dataBinding generator does convert those to camelCase and the standard for these xml files seems to be snake_case.

phillipthelen (changes requested):

Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/subscriptions/SubscriptionDetailsView.kt
phillipthelen: For the actual tests, we can also go over the test cases together to make sure we catch as many cases as possible.

Habitica/res/layout/subscription_details.xml
Hafizzle: Yeah can do!

Habitica/src/main/java/com/habitrpg/android/habitica/ui/views/subscriptions/SubscriptionDetailsView.kt
Hafizzle: For sure I can do that

Issue #1417, opened 2020-11-11
Improve Avatar Customization Buy Modals


saraolson:
Trying to unite the avatar customization experience with the normal Shop experience, starting with backgrounds. I've also included a preview step when you tap a background that isn't equipped, so you can see how it would look before committing (since you can scroll pretty far down on the background screen and may not see your avatar change easily).

Note that the bundle doesn't have the ability to pin.

<img width="1075" alt="Screen Shot 2022-04-21 at 3 09 23 PM" src="https://user-images.githubusercontent.com/12453119/164535487-0f8f327d-ff28-4f8d-89de-ba41886d921b.png">

[Zeplin link](zpl.io/vMLZ3XY)

PR #1772, opened 2022-05-15
#1417 - Improve Avatar customization Buy Modals

Hafizzle:
Show avatar model when purchasing background
Show avatar model & background when equipping a purchased background


Habitica/src/main/java/com/habitrpg/android/habitica/ui/AvatarView.kt
```diff
@@ -433,6 +437,22 @@ class AvatarView : FrameLayout {
         }
     }
 
+    fun setAvatarWithSelectedBackground(avatar: Avatar, backgroundForPurchase: String?) {
```
phillipthelen, line 440: Can we maybe make this more generic? Right now it only supports changing the background. If we decide to also show a preview for things like gear or other customisations, this would work anymore. Maybe extend the existing setAvatar method with an optional parameter `previews` (not sure if that really is the best name, but what I just came up with), that is a Map<LayerType, String>. That way we could replace each layer of the avatar if we want to.

Hafizzle: Got it got it that makes sense to me - I like previews so I'll be using that 👍

phillipthelen: The previews code looks good, but I think we don't actually need this method now anymore. Since enow that can be achieved using the previews one.

PR #1815, opened 2022-08-02
Show all to-dos and reset after cron

Hafizzle:
Save to-dos in sharedpref and reset after cron

### DIFF
--- a/wearos/src/main/java/com/habitrpg/wearos/habitica/ui/viewmodels/TaskListViewModel.kt
+++ b/wearos/src/main/java/com/habitrpg/wearos/habitica/ui/viewmodels/TaskListViewModel.kt
@@ -1,9 +1,13 @@
 package com.habitrpg.wearos.habitica.ui.viewmodels
 
+import android.content.SharedPreferences
+import androidx.core.content.edit
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.asLiveData
 import androidx.lifecycle.viewModelScope
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
 import com.habitrpg.common.habitica.models.responses.TaskDirection
 import com.habitrpg.common.habitica.models.responses.TaskScoringResult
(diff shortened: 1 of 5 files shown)


wearos/src/main/java/com/habitrpg/wearos/habitica/ui/viewmodels/TaskListViewModel.kt
```diff
@@ -65,4 +64,90 @@ class TaskListViewModel @Inject constructor(
...
+            } else {
+                savedCurrentTasks
+            }
+        }
+        return null
+    }
+
+    private fun mapTodos(tasks: List<Task>): List<Any>? {
+        saveCurrentToDos(tasks)
+        return getCurrentToDos()
```
phillipthelen, line 110: Is it necessary to do these two steps completely every time the task data updates? I think it should be sufficient to only load the task list from the SharedPreferences when the view model is created. And only save them when the view model is destroyed (I believe overriding `deinit` would be the right method for this). And all other operations during the lifetime of the view model are just done on a variable holding a list of the completed todos.

wearos/src/main/java/com/habitrpg/wearos/habitica/ui/viewmodels/TaskListViewModel.kt
```diff
@@ -65,4 +64,90 @@ class TaskListViewModel @Inject constructor(
...
+
+    private fun saveCurrentToDos(tasks: List<Task>) {
+        val taskList = mutableListOf<Task>()
+        val type: Type = object : TypeToken<ArrayList<Task?>?>() {}.type
+        if (tasksString != null) {
+            val savedCurrentTasks = gson.fromJson(tasksString, type) as MutableList<Task>
+            if (!savedCurrentTasks.isNullOrEmpty()) {
+                for (task in tasks) {
+                    if (!savedCurrentTasks.contains(task)) {
+                        taskList.add(task)
```
phillipthelen, line 121: I would only save completed todos and not all todos, since for incomplete todos we use the the repository as data source.

phillipthelen (approved):

Issue #1844, opened 2022-08-29
Audio plays at the volume of Notifications instead of Media volume


svenfulen:
**Describe the bug**
When you complete a task, level up, etc, anything that causes a sound to play, the sound is played out of the speaker as a notification sound rather than as a Media sound.

**To Reproduce**
Complete a habit positive or negative
 
**Expected behavior**
The sound should play at Media volume instead of at Notification volume.

**Screenshots**
![Screenshot_20220829-184204_Habitica](https://user-images.githubusercontent.com/73049985/187317721-f313bb5c-100b-4c69-99e1-d22277f7b934.jpg)

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S21
 - OS: Android 12
 - App Version: 4.0.1
 - Habitica Username: svenfulen

**Additional context**
I usually have my notifications turned up all the way, it is kind of jarring for an extremely loud sound to play every time something is checked off

PR #1847, opened 2022-08-31
Set audio to play at Media volume (Issue #1844)

svenfulen:
Set the AudioManager to use Music (media) volume instead of Notification volume

Habitica Username - svenfulen

my Habitica User-ID:


saraolson:
@svenfulen thanks for your patience as we worked your contribution into the update schedule! we've merged and tested it, and everything works well :) 

I've added this contribution to your Habitica profile and granted you a tier 🎊  thanks so much for helping!

Issue #1842, opened 2022-08-23
Show current HP in decimals if it's under 1


saraolson:
Right now if someone has 0.6 HP, android shows their hp as 0. They get confused wondering why they aren't running out of hp and getting reset to 50. In reality they haven't actually hit 0 HP yet. 

The decimal we show should be limited to the first number that isnt 0. ie if its 0.6, show that. if it's 0.06, show that.

booker-lee:
I'd like to help with this issue, let me know if it is ok for me to take it.

saraolson:
@booker-lee you're welcome to take this issue on! i've marked it as in progress for you. 

if anything comes up and you can't finish this issue, just let us know.

booker-lee:
@saraolson I pushed a change for this one. One thing to note is that if the value is lower than 0.1 (ex: 0.06), this fix would display 0.1 as the minimum hp. I feel like it should be enough for users and we don't have to deal with display of extreme low value (some thing like 0.00000001). Let me know if there are any suggestion for this PR.

saraolson:
@booker-lee that sounds good to me! our developer will reach out after he's had a chance to review if he has any suggestions

PR #1853, opened 2022-09-11
Fix hp display when value is under 1 #1842

booker-lee:
The cause of this issue comes from the ValueAnimator. The under 1 value is truncated due to the cast to int in ValueAnimator. Add animation for floating point values (when value is between 0 ~ 1)

my Habitica User-ID: c6558c97-ff5a-4a3f-8413-c562f907b40f


saraolson:
@booker-lee thanks for submitting this! we'll review when we're working on the next android release soon :)

booker-lee:
@saraolson, I'd like to know if this merged PR would be eligible for contributor tier progress. Thanks.

saraolson:
@booker-lee hi there! we usually add the PR to your account after its been merged and tested in the build to confirm we don't want any more changes. I'm still in the process of testing everything, but should be done shortly!

saraolson:
@booker-lee just finished lookin through it in the build, everything looks good! thanks for working on this :) i've added this contribution to your habitica profile and granted you a tier 🎊

PR #1862, opened 2022-10-11
Notes and checklist item link fixes

Hafizzle:
Allow edittext to be clickable for Notes and task checklist items (And prevent link from opening when clicking on open spaces)


phillipthelen:
is this for notes that are so long that they are not completely displayed in the list?
When the user saves their task, is the version saved _with_ the zero-width-space?

phillipthelen:
@Hafizzle apart from it potentially saving with the extra space, it looks good. Can you look into that?

Hafizzle:
> @Hafizzle apart from it potentially saving with the extra space, it looks good. Can you look into that?

Yeah! It does not save with the extra space - for my own sanity as well I tested to make sure the extra space is not being saved across notes and checklist items.

Issue #1846, opened 2022-08-30
display players class on profile


rilygreat:
> Device: Xiaomi Redmi 8A Dual
> Android Version: 29
> AppVersion: Version 4.0.1 (4460)
> User ID: 7747905d-9592-441c-8e10-df377041a795
> Level: 47
> Class: warrior
> Is in Inn: false
> Uses Costume: false
> Custom Day Start: 0
> Timezone Offset: -330
> Details:
> 
>  Greetings,
> I have been noticing something different lately on player's bio. I figured out that in the android version of the app, the player class symbol is not displayed near the player level. Hope the team will fix this issue.

saraolson:
we could easily add class after the player's level on their profile. the icon would be nice, but simply having 'Level 18 Rogue' would be enough for now

AdamWier:
Hello, I would like to work on this for Hacktoberfest. To be clear, by profile, you mean the screen that is displayed when you click on an avatar in your party? And the text 'Level XX <<Class>>" should be displayed where we currently see "Level XX" below the picture of their avatar?

saraolson:
@AdamWier thats right! i'll mark this issue as in progress for you. just let us know if you have to drop it for any reason :) thanks!

AdamWier:
I've posted a pull request with a bit of an explanation :)

PR #1865, opened 2022-10-16
If no flags loaded, assume character has class selected

AdamWier:
This is for isseu #1846

my Habitica User-ID: ef32aaa4-839b-480c-8218-9b609907d413

**Hello. First, please excuse me if there are any conventions for branch names or commits I didn't follow. I couldn't find a page on it.**

# Analysis of the problem
It seems like flags are only loaded for the current user. This means that when we load a profile page, there are no flags and the hasClass check fails. For the time being I have changed the method to assume true if there are no flags loaded. This displays "warrior" for those who don't have a class, which is behavior that is is consistent with the web version. 
I wasn't sure if this is the right correctio. Because: 
* if we should be loading the flags from the AP, the bug is with the API and not the mobile app
* But if we shouldn't receive flags, would it maybe be better to remove checks to hasClass since the user object assigned warrior, and as far as I can tell functions like one?

### DIFF
--- a/shared/src/commonMain/kotlin/com/habitrpg/shared/habitica/models/Avatar.kt
+++ b/shared/src/commonMain/kotlin/com/habitrpg/shared/habitica/models/Avatar.kt
@@ -27,11 +27,12 @@ interface Avatar {
 
     val hasClass: Boolean
         get() {
-            return preferences?.disableClasses != true && flags?.classSelected == true && stats?.habitClass?.isNotEmpty() == true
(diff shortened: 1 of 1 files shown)


saraolson:
@AdamWier thanks so much for lookin into this for us and submitting your PR! we've merged and reviewed it, and it looks good 🎊  the issue with showing class when a user is technically 'classless' is def a larger discrepancy we need to sync up on between platforms. but happy to have classes for accounts that have selected them show properly.

i've added this contribution to your habitica profile and granted you a tier :)

PR #1877, opened 2022-11-04
Android 13 Notification Permissions

Hafizzle:
- Added notification permissions
- Request permission on startup and when enabling push notifications

Permissions are requested only if user currently has not approved notification permission on Android 13.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/preferences/PreferencesFragment.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/preferences/PreferencesFragment.kt
@@ -3,6 +3,7 @@ package com.habitrpg.android.habitica.ui.fragments.preferences
 import android.content.Intent
 import android.content.SharedPreferences
 import android.content.res.Configuration
+import android.os.Build
 import android.os.Bundle
 import android.view.View
 import androidx.activity.result.contract.ActivityResultContracts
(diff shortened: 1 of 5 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/preferences/PreferencesFragment.kt
```diff
@@ -209,7 +214,11 @@ class PreferencesFragment : BasePreferencesFragment(), SharedPreferences.OnShare
                 pushNotificationsPreference?.isEnabled = usePushNotifications
                 userRepository.updateUser("preferences.pushNotifications.unsubscribeFromAll", !usePushNotifications).subscribe()
                 if (usePushNotifications) {
-                    pushNotificationManager.addPushDeviceUsingStoredToken()
+                    if (!pushNotificationManager.notificationPermissionEnabled() && (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU)) {
```
phillipthelen, line 217: Is it actually necessary to check for tiramisu? Since that is the only version where the user would even need to approve notifications, the `notificationPermissionEnabled` call should only return `false` on devices with android 13 or higher. Right?

saraolson:
if notifications are off on a system level and i go to settings, push notifs appear still checked and if i uncheck/recheck them i'm not given a prompt to allow notifs. i also added a reminder and wasn't prompted to turn them on.

when notifications were turned off and i launched the app initially i got a prompt, which was expected. i said no to that prompt then made sure it wouldnt spam me with it and all that worked as expected. i just expected the prompt to come up again when i perform an action that needs notifs on.

PR #1940, opened 2023-03-22
Updated find more members description

Hafizzle:


phillipthelen:
This looks like it just replaces the title text with the empty version? Did something during merging maybe go wrong?

PR #1949, opened 2023-03-29
Default Dailies to Due for new users

Hafizzle:


Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/tasks/TaskRecyclerViewFragment.kt
```diff
@@ -537,7 +538,14 @@ open class TaskRecyclerViewFragment :
     }
 
     private fun setPreferenceTaskFilters() {
-        (activity as? MainActivity)?.viewModel?.user?.observeOnce(this) {
+        val isNewUser = mainActivity?.viewModel?.isNewlyCreatedUser
+        if (isNewUser == true) {
+            lifecycleScope.launchCatching {
+                userRepository.updateUser("preferences.dailyDueDefaultView", true)
```
phillipthelen, line 544: Instead of setting this to true here fore new users, we should update the preference to match what the task filter was updated to.
So for all users it should check if the filter that was changed is that for dailies, it should update the preference accordingly, so that on next app launch it launches again with the same filter

PR #1954, opened 2023-04-06
Update state / message shown when you have all armoire items

Hafizzle:


phillipthelen:
Doesn't this remove the message that should be shown when you don't have all items yet?

PR #1956, opened 2023-04-11
Use InsufficientGemsDialogEntryPoint

Hafizzle:


phillipthelen:
Hilt has `@EntryPoint` annotations for this use case. Can you change the PR to use that? it is used in the `PetSuggestHatchDialog`.

There are two parts needed. The annotated interface
```
@EntryPoint
    @InstallIn(SingletonComponent::class)
    interface PetSuggestHatchDialogEntryPoint {
        fun useCase(): HatchPetUseCase
    }
```
And then in the init method requesting and setting the fields
```
        val hiltEntryPoint = EntryPointAccessors.fromApplication(context, PetSuggestHatchDialogEntryPoint::class.java)
        hatchPetUseCase = hiltEntryPoint.useCase()
```

It's not super well documented how to implement this, but taking the `PetSuggestHatchDialog` implementation as a base should help, since in the end it's relatively simple!

PR #1964, opened 2023-05-04
Find more members w/ header and displayname styling fix

Hafizzle:


Habitica/src/main/java/com/habitrpg/android/habitica/ui/fragments/social/party/PartySeekingFragment.kt
```diff
@@ -178,12 +179,14 @@ fun PartySeekingView(
                         .fillMaxWidth()
                         .padding(top = 22.dp, bottom = 14.dp)
                 ) {
-                    Text(
-                        stringResource(R.string.find_more_members),
-                        color = HabiticaTheme.colors.textPrimary,
-                        fontSize = 16.sp,
-                        fontWeight = FontWeight.Medium
-                    )
+                    Box(modifier = Modifier.background(Color.Transparent)) {
```
phillipthelen, line 182: How does this change how it is displayed? Doesn't it just wrap it in an invisible box of the same size meaning it looks/behaves the same?

PR #1982, opened 2023-05-19
Change class flow from settings update

Hafizzle:
- tap change your class in settings
- get a confirmation modal with a red button to confirm this will cost 3 gems (you aren't charged at this point)
- view the class options
- select your class and get another confirmation modal
- confirm, you're charged, and your class changes


saraolson:
2. when i tap change class from settings without enough gems, i get a warning that i don't have enough but it leaves me on the class selection screen. we should probably just check that gem balance is above 3 as soon as they tap the entry from settings.
3. the not enough gems modal that shows is the basic one, we'd want to show the fancy one that allows them to purchase gems right there
4. if i try to select a class, i get two warning modals back to back
5. i backed out and was left without a class and the enable class system button doesn't work, so i am stuck

saraolson:
testing 7/7:
- when i leveled up to 10 and tapped 'Become a Rogue' it gave me a modal asking if i want to change class for 3 gems. it didnt actually charge me for them, so i think that modal string just needs to be updated to the new design that doesn't mention 3 gems. see below

PR #1989, opened 2023-05-26
Report User Feature

Hafizzle:


phillipthelen:
Does the server accept the flagging call like this? With the text instead of the message ID?

PR #1996, opened 2023-06-15
Moderator Option Fixes

Hafizzle:


Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.kt
```diff
@@ -125,38 +121,32 @@ class FullProfileActivity : BaseActivity() {
...
                     binding.adminStatusView.isVisible = isModerator
-                    if (isModerator) {
-                        val member = socialRepository.retrieveMember(userID, true)
-                        member?.stats = this@FullProfileActivity.member.value?.stats
-                        if (member != null) {
-                            updateView(member)
-                        }
+
+                    //Get Current viewed profile user
+                    val member = socialRepository.retrieveMember(userID)
```
phillipthelen, line 142: We need to do two calls here for moderators. This returns the actual user data. But to display if a user is shadow muted or not, it also need to do the `socialRepository.retrieveMember(userID, true)` call to get the data from hall of heroes. Unfortunately that does not contain all info about the member. So we'd need to change it, so that the data returned from the hall of heroes call is only used to set the adminStatusView

phillipthelen (changes requested):

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/FullProfileActivity.kt
Hafizzle: tyvm! That's super helpful - I'll make those changes

PR #1999, opened 2023-06-20
Death Screen update to handle small screens or large text

Hafizzle:
- NestedScrollView added to handle smaller phone screen sizes
- Set Buttons and Broken equipment text relative to bottom of screen (versus below text)

### DIFF
--- a/Habitica/res/layout/activity_death.xml
+++ b/Habitica/res/layout/activity_death.xml
@@ -1,91 +1,119 @@
 <?xml version="1.0" encoding="utf-8"?>
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


saraolson:
issues when testing this:
- the text was cut off and made to scroll, rather than the whole view. would like all the text to remain visible and the view to move
- i got a server error when tryin to press the button and it didnt work, then just left me on my task screen again. not sure what happened there?
-the ghost and heart location got messed up, where the ghost got pushed too far down, making its tail stick out from the heart and hiding it's face. would want those elements to scale and move together (though this is lower priority)


https://github.com/HabitRPG/habitica-android/assets/12453119/0a05577b-8360-4957-adda-c1f1236fcc2d

saraolson:
^ still the case when testing 7/7 (moved back to in progress)

Hafizzle:
@saraolson should now be fixed with f5159e1186cf499249221fca81a76777bd303459

PR #2004, opened 2023-06-26
Open Party chat when pressing on Chat notification

Hafizzle:


saraolson:
the in-app notification opens chat, but the push notification still opens the details tab of the party for me

Hafizzle:
@saraolson should now be fixed with c9e9d4684dea92e0d1498dafcb12c7f5b82dfecf

PR #2007, opened 2023-06-30
Match web implementation for tags / order tags

Hafizzle:
- Order tags
- Add "group" field to Tag


saraolson:
some suggestions:
- Change category labels to 'Group Tags' / 'Challenge Tags' / 'Your Tags'
- Show the above categories in the filter sheet as well

PR #2010, opened 2023-07-05
Handle non-HTTPExceptions & continue flow if consume fails & no error

Hafizzle:
- Throw exception on consume if failed to consume (to continue flow/handle error)
- Handle other potential errors (such as IOException or a exception thrown by billingClient.consumePurchase method that would not previously be handled


saraolson:
hopefully fixed in 4.2.3

PR #2015, opened 2023-07-12
Add tag category titles to filters

Hafizzle:
- Add tag category titles to filters
- Update tag title strings


phillipthelen:
Can you take out the unrelated changes/reverts? Then I think it's ready to be merged

PR #2026, opened 2023-08-28
Review Manager - In App Review

Hafizzle:
In-App Review manager logic -

Initial Check:
When determing to request a in-app review, first check if an initial review request has been made.
If no review has been requested before, the app takes note of the current number of user check-ins and proceeds to request a review on the next startup during the next logical break.

Logical Breaks:
After the initial in-app review request, wait for a logical break before potentially asking the user again. This is to ensure we don't pester them :)

Limit on Total Review Requests:
Request a review only up to 5 times. After that, it will not prompt the player for reviews anymore (Unless a user has checked-in 75 times before requesting 5 in-app reviews).

Check-in Limit Since First Review:
If a user has checked in more than 75 times since the very first review was requested, the app won't request another review. This is to ensure that players are not continuously prompted.

Gap Between Review Requests:
Once a review has been requested, wait until the player has checked in at least 5 more times before considering another review request. This ensures a decent gap between consecutive review prompts.

### DIFF
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
@@ -12,19 +12,19 @@ class ReviewManager(private val context: Context) {
     companion object {
         private const val REVIEW_REQUEST_COUNT_KEY = "ReviewRequestCount"
         private const val INITIAL_CHECKINS_KEY = "InitialCheckins"
-        private const val REVIEW_REQUESTED = "ReviewRequested"
+        private const val SHOULD_QUEUE_REVIEW = "ShouldQueueReview"
         private const val LAST_REVIEW_CHECKIN_KEY = "LastReviewCheckin"
     }
 
     private fun canRequestReview(currentCheckins: Int): Boolean {
         val initialCheckins = sharedPref.getInt(INITIAL_CHECKINS_KEY, -1)
-        val reviewRequested = sharedPref.getBoolean(REVIEW_REQUESTED, false)
+        val shouldQueueReview = sharedPref.getBoolean(SHOULD_QUEUE_REVIEW, false)
         val lastReviewCheckin = sharedPref.getInt(LAST_REVIEW_CHECKIN_KEY, -1)
 
-        if (reviewRequested) {
+        if (!shouldQueueReview) {
             // First review request has been made, wait for following request (if any)
             // to request again in the spirit of asking during a logical break.
-            sharedPref.edit().putBoolean(REVIEW_REQUESTED, true).apply()
+            sharedPref.edit().putBoolean(SHOULD_QUEUE_REVIEW, true).apply()
             return false
         }
 
--- a/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
+++ b/Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
@@ -1,0 +1,34 @@
+package com.habitrpg.android.habitica.helpers
+
+import android.content.Context
+import androidx.appcompat.app.AppCompatActivity
+import com.google.android.play.core.review.ReviewManagerFactory
+
+class ReviewManager(private val context: Context) {
+
+    private val reviewManager = ReviewManagerFactory.create(context)
+    private val sharedPref = context.getSharedPreferences("ReviewPrefs", Context.MODE_PRIVATE)
+
+    fun hasShownReviewForEvent(eventKey: String): Boolean {
+        return sharedPref.getBoolean(eventKey, false)
+    }
+
+    fun setReviewShownForEvent(eventKey: String) {
+        sharedPref.edit().putBoolean(eventKey, true).apply()
+    }
+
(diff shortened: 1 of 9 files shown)


Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    private fun canRequestReview(currentCheckins: Int): Boolean {
+        val initialCheckins = sharedPref.getInt(INITIAL_CHECKINS_KEY, -1)
+        val shouldQueueReview = sharedPref.getBoolean(SHOULD_QUEUE_REVIEW, false)
+        val lastReviewCheckin = sharedPref.getInt(LAST_REVIEW_CHECKIN_KEY, -1)
+
+        if (!shouldQueueReview) {
+            // First review request has been made, wait for following request (if any)
+            // to request again in the spirit of asking during a logical break.
+            sharedPref.edit().putBoolean(SHOULD_QUEUE_REVIEW, true).apply()
```
phillipthelen, line 27: Should use kotlins edit {} shorthand/function

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
```diff
@@ -0,0 +1,81 @@
...
+            return false
+        }
+
+        if (initialCheckins == -1) {
+            // Store the current checkins as the initial value
+            sharedPref.edit().putInt(INITIAL_CHECKINS_KEY, currentCheckins).apply()
```
phillipthelen, line 33: Should use kotlins edit {} shorthand/function

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
```diff
@@ -0,0 +1,81 @@
...
+                val flow = reviewManager.launchReviewFlow(activity, reviewInfo)
+                flow.addOnCompleteListener { _ ->
+                    // The flow has finished - Increase the request count.
+                    incrementReviewRequestCount()
+                }
+            }
+        }
+
+        // Save the current checkins after a successful review request
+        sharedPref.edit().putInt(LAST_REVIEW_CHECKIN_KEY, currentCheckins).apply()
```
phillipthelen, line 73: Should use kotlins edit {} shorthand/function

Habitica/src/main/java/com/habitrpg/android/habitica/helpers/ReviewManager.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private fun incrementReviewRequestCount() {
+        val currentCount = sharedPref.getInt(REVIEW_REQUEST_COUNT_KEY, 0)
+        sharedPref.edit().putInt(REVIEW_REQUEST_COUNT_KEY, currentCount + 1).apply()
```
phillipthelen, line 78: Should use kotlins edit {} shorthand/function

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/ArmoireActivity.kt
```diff
@@ -114,6 +118,15 @@ class ArmoireActivity : BaseActivity() {
             val args = ArmoireActivityArgs.fromBundle(it)
             equipmentKey = args.key
             configure(args.type, args.key, args.text, args.value)
+
+            if (args.type == "gear") {
+                userViewModel.user.observeOnce(this) { user ->
+                    val totalCheckIns = user?.loginIncentives
+                    if (totalCheckIns != null) {
```
phillipthelen, line 125:
```
user?.loginIncentives?.let { totalCheckins ->

}
```

This would be a slightly more concise/kotlin version

Habitica/src/main/java/com/habitrpg/android/habitica/ui/activities/ClassSelectionActivity.kt
```diff
@@ -292,4 +300,11 @@ class ClassSelectionActivity : BaseActivity() {
             finish()
         }
     }
+
+    private fun checkForReviewPromptAfterClassSelection() {
+        userViewModel.user.observeOnce(this) { user ->
+            val totalCheckIns = user?.loginIncentives ?: return@observeOnce
```
phillipthelen, line 306: This version also works (see my other comment. But I think we should stick to one way.

phillipthelen (changes requested):
Left some comments. But looks good so far!

We should also change it to use Hilt for the dependency injection.
Steps for that:

1. add @Inject to the `latent var reviewManager: ReviewManager` lines 
2. Remove the initialisations
3. add the following code to the [AppModule.kt](https://github.com/HabitRPG/habitica-android/blob/main/Habitica/src/main/java/com/habitrpg/android/habitica/modules/AppModule.kt) file

```
@Provides
    fun providesReviewManager(@ApplicationContext context: Context): ReviewManager {
        return ReviewManager(context)
    }
```

That should already be enough to get the ReviewManager available through hilt

Issue #2031, opened 2023-09-04
Daily notification bug


pwall2222:
**Describe the bug**
`Repeats Daily` tasks get a notification every day. Even if `Every` is set to more than 1.

**To Reproduce**
- Create a task in `Dailies` section.
`Scheduling`:
-  Set `Repeats` to `Daily`.
- Set `Every` to more than 1 (`>1`).
Finally:
- Add a reminder.

**Expected behavior**
Getting a notification only on the days that the task is active.

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A7 2018
 - OS: Android 10 One UI 2.0
 - App Version: 4.2.5
 - Habitica Username: @pwall2222

**Additional context**
Happend on my Android 11? 12? device aswell it's now broken so can't do any testing on that.

pwall2222:
> @pwall2222 if you delete that reminder, then add a new reminder (so basically, adding a reminder after the scheduling change has previously been saved) does that fix the issue?

I have tested removing the reminder and even the task and saving without reminder and then adding it. Doesn't fix the issue.

pwall2222:
I think this specific bug is solved with last release!

Issue #1972, opened 2023-05-10
Disable upcoming reminder if task is alredy done 


pwall2222:
When you create a task and create this reminders:
- In 30 min
- In 1 hour
- In 2 hour
And you after 1 hour check off the task, the 2 hours reminder will still pop up even though the task is finished.
(From what I have searched through the code, it's a matter of when completing a task in completeTaskForUser, to delete or disable the reminders)

phillipthelen:
The app checks if a task was completed before displaying the notification. Did you complete your task in the app or on the website? Otherwise that check might not be working correctly in your case.

pwall2222:
I'm checking it on my device, and I go look and the task is checked off alredy too but I still get the notification.

saraolson:
haven't been able to reproduce on 4.2

saraolson:
had a report of this from a user on to do reminders, which may be why we couldnt reproduce? we were testing daily reminders. will test this.

PR #2039, opened 2023-10-24
Add Task model tests

danielgomezrico:

This PR will add unit tests to the `Task` model, I saw some issues related to reminders and dates and wanted to add tests to be sure that this logic is working as expected.

Related issues:
- https://github.com/HabitRPG/habitica-android/issues/2031
- https://github.com/HabitRPG/habitica-android/issues/1972

Are there any other cases that I should be covering here?

Can you add `hacktoberfest-accepted` label to this PR please?

### DIFF
--- a/Habitica/src/test/java/com/habitrpg/android/habitica/models/TaskTest.kt
+++ b/Habitica/src/test/java/com/habitrpg/android/habitica/models/TaskTest.kt
@@ -1,0 +1,63 @@
+package com.habitrpg.android.habitica.models
+
+import com.habitrpg.android.habitica.models.tasks.Task
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+import java.util.Date
+import java.time.ZonedDateTime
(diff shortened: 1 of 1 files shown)


danielgomezrico:
Tests are not passing but its related to main codebase, trying to fix them on https://github.com/HabitRPG/habitica-android/pull/2045

Hafizzle:
Hey @danielgomezrico,

First off, big thanks for diving into Task model tests! It's really cool to see your initiative in adding unit tests, especially around reminders and dates. This is super important for keeping everything running smoothly.

I've gone through your tests, and they're a solid start – they cover the basic scenarios well. However, we're thinking it might be even better to add some more real-world complexity to these tests. For example:

How about testing how the isDueToday function behaves across different time zones?
Maybe throw in some scenarios for leap years or unusual date setups?
Testing the function with a variety of task types and configurations could also be really insightful.

These additions would really beef up the robustness of our tests and make sure we're covering as many user scenarios as possible.

Thanks again for your work :)

PR #2052, opened 2023-12-17
Fixes #1861: Make the entire like layout clickable vs just the text

### LINKED ISSUE #1861: Improve tap area on liking messages
Sometimes it can be difficult to tap the right area to like a message, we should take a look at that tap area and see if we can make it easier to tap.

jsoberg:
Previously just the text itself was clickable here, so I made the parent layout clickable instead. Ideally this layout would be a bit larger for a11y reasons, Google recommends 48dp minimum touch target size ([link](https://support.google.com/accessibility/android/answer/7101858?hl=en)) but I left the sizing as-is for now.

my Habitica User-ID: `caef89d8-3a3d-4d3b-a5ad-67cb0455a56e`

### DIFF
--- a/Habitica/res/layout/chat_item.xml
+++ b/Habitica/res/layout/chat_item.xml
@@ -72,15 +72,18 @@
                         android:layout_height="wrap_content"
                         android:layout_gravity="center_vertical"
                         android:layout_marginStart="5dp"
+                        android:clickable="true"
+                        android:focusable="true"
                         android:background="@drawable/layout_rounded_bg">
+
                         <TextView
                             android:id="@+id/tvLikes"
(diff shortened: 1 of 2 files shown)


Hafizzle:
Hey @jsoberg, thanks a bunch for this tweak! 

Really smart move to make the entire layout clickable – it's little changes like these that can make a big difference, especially for accessibility. 

As for the touch target size - good call! I think keeping Google's guidelines in mind is super important. We might play around with the sizing down the line, but for now, what you've done is spot on. TY!

jsoberg:
Thanks @Hafizzle! I love the app btw, I'm hoping to contribute more regularly as time permits!

PR #2078, opened 2024-07-10
if else shorter

tarat80:
android studio add parameter names and I shorted if else branch. 
I am not sure, but is that activity must build character classes to show them on screen. There enough put static pictures. And after class selection activity check changes locally in mainuserviewmodel, while changes was remote on server. MainUserViewmodel not  source of truth.
my Habitica User-ID:
83ef547b-4ca4-4ace-8244-92dd044fbcac


tarat80:
i did better

PR #2093, opened 2024-11-13
Improve Gradle Configuration

### LINKED ISSUE #2075: [Suggestion] Update project Gradle setup (move to version catalogs, add shared build-logic)
Hey all! I have a few improvement suggestions that I wouldn't mind opening some PRs for surrounding the Gradle build logic setup for the project:

1. Move to using version catalogs ([Gradle docs](https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml) and [Android docs](https://developer.android.com/build/migrate-to-catalogs)) for dependencies. This provides a central file for defining dependencies/versions, making them accessible in a type-safe manner (through `libs.X`).

2. Create a new module for storing shared build logic ([example](https://github.com/android/nowinandroid/tree/main/build-logic)). I noticed some duplicate code in build files (one example being adding build config fields in [wearOS](https://github.com/HabitRPG/habitica-android/blob/main/wearos/build.gradle#L188-L204) and [Habitica](https://github.com/HabitRPG/habitica-android/blob/main/Habitica/build.gradle#L331-L347)) which would be a good first step into moving to a shared function that any build script in the project could use. In the future, this could be used to create convention plugins and other shared build logic/setup between modules as desired.

Let me know if you think these additions would be worthwhile for the project and I'll start working on PRs for them.

jimlyas:
This PR include:

- Migrating Groovy DSL into Kotlin DSL
- Migrating from dependency management based on extras to version catalog
- Migrating from KAPT to KSP (Realm still use KAPT because current version required it)
- Enable parallel and build caching in gradle properties
- Enable typesafe project accessors
- Change allUnitTest task's type from GradleBuild into DefaultTask
- Move share build logic into plugins

Should resolve #2075 

my Habitica User-ID: 7b1043d4-0272-410c-9c04-721b667f1566


phillipthelen:
This looks great! you mention KSP not working for realm yet: does that mean the PR is not ready for review/testing and merging or does realm use KAPT for now?

jimlyas:
What I mean is KSP is already implemented for moshi and hilt.

Current realm gradle plugin require the project to use kapt. Might work if we migrate to `io.realm.kotlin` and refactor the code. But it might be too far-fecthed from what I originally intended.

This PR still use kapt only for realm.

But the PR is ready for testing and review. 

I notice the pipeline is failing because there's no signingrelease.properties, in my local build it's running because I rename the template file. Did I miss something? 

Thanks in advance.

jimlyas:
I already fixed the pipeline issue, previously the build script always looking for the signingrelease.properties

phillipthelen:
Thank you for all this work. Wanted to do this for a while and never got to it.

PR #2102, opened 2025-01-25
[Improvement] Apply top margins to notification prompt views in create habit/daily screen

jsoberg:
I noticed that the notification prompts on the create habits/dailies screen were touching the views above them. This PR just adds the same top margin that was being applied as bottom margins, so that the notification prompts have some space.

| Before | After |
| ----- | ----- |
|  |  |
|  |  |

my Habitica User-ID: caef89d8-3a3d-4d3b-a5ad-67cb0455a56e


phillipthelen:
Thanks for this! I'll close this issue, since the fix here is to not show this prompt for habits, since they don't have reminders. But this caused me to be aware of the issue! I'll fix it showing in the wrong cases.

PR #2126, opened 2025-05-31
Improve Talkback for tiles

### LINKED ISSUE #1874: TalkBack issues regarding purchase menus and feedback item equipping
> Device: samsung Galaxy S21 5G
> Android Version: 31
> AppVersion: Version 4.0.4 (4651)
> User ID: b859efbd-0073-4a8c-8c2a-f768dbec676a
> Level: 2
> Class: warrior
> Is in Inn: false
> Uses Costume: false
> Custom Day Start: 3
> Timezone Offset: -60
> Details:
> 
> Hi, I just recently started using your app and it's pretty good. I'm blind so I use TalkBack and you have done a great job at making the app accessible. However, the one problem I have encountered is in any of the menus where you can buy things. All it reads is a number, which I assume is the cost of the item. It gives me no feedback for what the item actually is. Also, in the equipment menu it reads me all of the different slots but it does not give me feedback for what I have equipped unless I click on the individual slot and find the item in the list.

koboldblue:
Changes made:
- Shop items now read out name of item, price with currency, limited availability, items owned and locked state (previously just price)
- In equipment, name of item or none will be read out in addition to slot (previously just slot)
- In stable, hatchable pets will be read out as name of pet and "Hatch Pet" (previously unlabelled)
- In task creation, the click listener for the difficulty has been moved to the surrounding column so icon and text are grouped (previously icon was read as unlabeled and text was not clickable)
- In navigation drawer, the profile picture is read out as "Profile" (previously unlabelled)

Fixes #1874

My Habitica User-ID: 2bcd10dc-0ea9-4d95-b7a6-f55fa07af537

### DIFF
--- a/Habitica/res/layout/drawer_main.xml
+++ b/Habitica/res/layout/drawer_main.xml
@@ -33,7 +33,8 @@
                     android:id="@+id/avatarView"
                     android:layout_width="70dp"
                     android:layout_height="70dp"
-                    android:layout_gravity="center" />
+                    android:layout_gravity="center"
+                    android:contentDescription="@string/profile"/>
         </com.habitrpg.android.habitica.ui.RoundedFrameLayout>
 
         <LinearLayout
--- a/Habitica/res/layout/row_shopitem.xml
+++ b/Habitica/res/layout/row_shopitem.xml
@@ -47,6 +47,7 @@
                 android:textColor="@color/text_quad"
                 android:paddingStart="2dp"
                 android:paddingEnd="2dp"
+                android:accessibilityTraversalAfter="@id/container"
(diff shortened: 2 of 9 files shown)


koboldblue:
Sorry, I just realised after installing the latest update and checking it on my personal phone that not all the talkback changes I made work on Samsung. If that's ok I would take another look later this week.

koboldblue:
@saraolson Sorry to be a bother about this, but what about contributor tier credit?

koboldblue:
@Hafizzle sorry to keep bringing this up, but I never heard back about contributor tier credit for this.

PR #2129, opened 2025-06-03
Removes task alarms/reminders before updating a task

Hafizzle:
any existing alarms/reminder associated with a task are cancelled before the task is updated. This prevents duplicate or conflicting alarms from being set.


Hafizzle:
Currently testing the updated alarm removal and scheduling logic, making sure that deleted reminders no longer trigger. (and confirming that unchanged tasks don’t create duplicate alarms)

phillipthelen:
merged this in manually

PR #2134, opened 2025-06-04
Skill dialog bottom sheet

Hafizzle:
Bottom sheet dialog for displaying skill information
Includes a new composable view: SkillDialog


phillipthelen:
merged this in manually

PR #2137, opened 2025-06-05
Fixes incorrect week calculation for monthly repeats

Hafizzle:
Using the built-in "week of month" value meant that any days before the 1st counted as “week 1,” so selecting the first Monday sometimes grabbed a date that actually belonged to a partial week from the previous month.

The new approach replaces that by taking the numeric day (1-31) and grouping 1-7 as week 1, 8–14 as week 2, and so on - so "first Monday" really means the first Monday falling in days 1–7 of that month, not a stray day from a partial week.


Hafizzle:
Currently testing

phillipthelen (commented):
Is this the same logic as used on web? If so, this looks good

PR #2155, opened 2025-06-18
postpone consumePurchase until after server-side purchases credit

Hafizzle:
- In PurchaseHandler.handleError, remove automatic consumePurchase on any error to avoid burning tokens before gems arrive.
- Only call consumePurchase() after the backend confirms gems have been credited.
- Add retry logic around startListening/queryPurchases so any unconsumed purchases are picked up when the app launches.
- Prevent lost purchases by ensuring the client only consumes tokens once we know the user got their gems.

(This change works in tandem with the backend fix that defers marking receipts consumed until after successful gem credit.)


Hafizzle:
Closing this - resubmitting another PR that handles this more cleanly

PR #2171, opened 2025-10-07
Feature/unifiedpush gate

sudoxreboot:
habitica app with self host and unified push


phillipthelen:
Hi,

Thank you for the work, however this is unfortunately not something we are planning to merge in at this moment due to increased code complexity and app size.

PR #2174, opened 2025-11-03
 Enable changing server address at runtime (Self-hosted support)

schrauger:
I updated the PR #2166 to include the changes requested by @Hafizzle. Using the code from the pr by @terofeev, I updated the toast message to only show up after the user has clicked 4 times, and the total clicks is increased to 7.

The custom server preference box also includes a warning at the top, notifying users that the accounts for custom servers are separate from the default server.

The "Base URL" text was also changed to "Default".

Note that all of the changes were created by @terofeev, except for the tweaks mentioned above (in `ServerSettingsDialog.kt` and `SequentialClickBox.kt`).


Hafizzle:
@schrauger Thanks for this PR! This looks like solid work. We're currently focused on some things in the pipeline, but I want to give this a proper look when we have more time. I'll circle back to review it thoroughly.

Really appreciate you taking the time to contribute!

titilambert:
Hello, any news for this PR ?

Issue #2180, opened 2025-12-15
task reminder time zone implicit set


r1ckdu:
**Describe the bug**
Time zone is applied silently when setting a reminder for a task.

**To Reproduce**
Go to other time zone country, set up reminders, come back to your home and be disappointeds
 the reminders don't fire on time.

**Expected behavior**
Reminders have timezones that can be set (defaulting to your local tz), or they are timezones independent.

**Smartphone (please complete the following information):**
 - Device: Fairphone 5
 - OS: Android 15
 - App Version: 4.9
 - Habitica Username: joerg

ThirFir:
Hi, I'm interested in this issue and would like to fix it. Could you please assign it to me?

saraolson:
This needs a little internal discussion from our team to be sure we implement this in a way we're happy with across both apps. Removing the help wanted label and adding the investigation label.

PR #2182, opened 2026-01-27
Fixes HabitRPG/habitica-android#2180

melissa-b-02:
# Fix

Reminders are now stored and handled as time zone independent, ensuring they always fire at the intended local time regardless of travel or system time zone changes.


Hafizzle:
TY for looking into this, we're going to handle this one on our end though.

Reminder times come back from the API with a UTC offset, and LocalDateTime.parse with ISO_LOCAL_DATE_TIME throws on those. the underlying timzone issue also needs to be worked on across web/ios/android simultaneously, so it's unfort not something we can fix on android side alone.

ty again for the findings, closing this out.

